PR #2, opened 2015-06-06
OS X support almost done. Fixed a crash using Glib::Dispatcher, also …

eidheim:
Actually, Glib::Dispatcher only made the code simpler and resulted in correct syntax highlighting, but the "std::min(buffer()->get_text().size(), raw_size)" fixed a crash.

### DIFF
--- a/juci/CMakeLists.txt
+++ b/juci/CMakeLists.txt
@@ -88,6 +88,14 @@ else()
   message("Gtkmm link dirs ${GTKMM_LIBRARIES}")
   message(FATAL_ERROR "The gtkmm libraries are required. Quitting.")
 endif()
+
+pkg_check_modules(GTKSVMM gtksourceviewmm-3.0)
+if(${GTKSVMM_FOUND})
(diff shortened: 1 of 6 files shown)


zalox:
I was thinking about a solution for this. I was thinking about creating a event manager for the entire project, for instance making certain objects that use threading inherit some sigc type so messaging is easier. But i quess Dispatcher works in a similar way?

eidheim:
Yes as I understand it the Glib::Dispatcher() merely tells the gui-main-thread to queue this function, much like the signals are queued (for instance signal_end_user_action()). This should be the correct way to program multithreaded applications in GTKMM now. Signals and Glib::Dispatcher() are, in this way, certain not to execute in parallell as long as the Glib::Dispatcher object is created in the main thread.

eidheim:
Totally forgot about the line numbers already in place. Activated line numbering in gtksourceviewmm instead, but the previous line numbering should be removed (I don't see this in OS X).

eidheim:
That last commit might be abit ugly. Hopefully nothing important got deleted!

PR #7, opened 2015-06-13
Improved indentention and some fixes

eidheim:
There are alot of really simple getters and setters, these should be removed at some point. Direct access to class members should be allowed, and this will result in smaller and simpler code. Bjarne Stroustrup agrees with this btw:)

Some cleanup should also be done with respect to model, view, controller as well. Not entirely sure how best to do this yet however. I could try to clean up source.h and source.cc maybe, and we could use this as a base for the rest of the classes. This would include getting rid of the simple getter and setter methods in these classes (source.h). 

I will also try to find the reason behind the segmentation fault when you open/close tabs several times.

### DIFF
--- a/juci/config.cc
+++ b/juci/config.cc
@@ -15,6 +15,9 @@ MainConfig::MainConfig() :
 void MainConfig::GenerateSource() {
   DEBUG("Fetching source cfg");
   boost::property_tree::ptree source_json = cfg_.get_child("source");
+  source_cfg_.tab_size=source_json.get<unsigned>("tab_size");
+  for(unsigned c=0;c<source_cfg_.tab_size;c++)
+    source_cfg_.tab+=" ";
   boost::property_tree::ptree syntax_json = source_json.get_child("syntax");
   boost::property_tree::ptree colors_json = source_json.get_child("colors");
   boost::property_tree::ptree extensions_json = source_json.get_child("extensions");
--- a/juci/config.json
+++ b/juci/config.json
@@ -28,7 +28,8 @@
 	    "hpp",
 	    "hxx",
 	    "h++"
-        ]
+        ],
+	"tab_size": 2
     },
     "keybindings": {
         "new_file": "<control>n",
--- a/juci/juci.cc
+++ b/juci/juci.cc
@@ -13,6 +13,7 @@ int main(int argc, char *argv[]) {
                                                                 argc,
                                                                 argv,
                                                                 "no.sout.juci");
+  Gsv::init();
   init_logging();
   Window window;
   return app->run(window);
--- a/juci/notebook.cc
+++ b/juci/notebook.cc
@@ -1,7 +1,6 @@
 #include <fstream>
 #include "notebook.h"
 #include "logging.h"
-#include <regex>
 
 Notebook::Model::Model() {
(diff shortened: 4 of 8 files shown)


eidheim:
The segmentation fault when opening/closing tabs several times is now fixed.

eidheim:
Not much left now before one can start develop Juci++ with Juci++! Some more smart indentation and of course search, and we might be there.

zalox:
I merged this and added some configurations. Background color etc can now be set in config.json. I also moved the Gsv::init to source::view constructor.

I'm having problems with line numbering and indention on current debian stable. Line numbers are not shown and indention works sometimes. In Debian testing there was some support. This needs to be investigated.

eidheim:
What version of gcc do you have in debian stable? One needs g++-4.9 for regex support. One solution is to use boost::regex, but gcc-4.9 should become standard in most distributions soon. 

The indentation is more of a temporary solution just to be able to write code in Juci++ without the need of too many key-strokes. It is currently very simple, but works ok in most of the cases. A better solution would perhaps be to search for the first open "{" and perform indentation based on this, and also search for open "(" and "[" (from the cursor to the first open "{") to perform appropriate indentation when functions and the like are on several lines. I can create this module when I got the time. It should not require too much work. However, why line numbers do not work in debian stable I do not know. Older gtk* versions?

Both line numbers and indentation worked in my Debian testing.

zalox:
I will check versions when I get home. It could be because I installed the sourceview from git instead of aptitude. I will investigate by running a fresh install of Debian on my virtual machine.

zalox:
I installed a new Debian stable box. It supported the following of the new features
* line numbers
* highlighting of current line

There was no indention support. I haven't had time to investigate more. I should make a clone of this box and try to run it in testing later.

eidheim:
What is the installed gcc version on debian stable?

zalox:
https://packages.debian.org/jessie/gcc

eidheim:
Just to be sure, what does this say: "g++ -v"?

zalox:
I'm not at the same computer right now, but it also got debian stable. I tested support and I got the same result as at home.
```bash
gcc version 4.9.2 (Debian 4.9.2-10) 
```
I opened notebook.cc and edited line 274 and tried to write an if-statement. No indention on any use cases.

eidheim:
Ok, I'll install debian stable on a VM and check.

zalox:
Ok, could you try with this command?
```bash
sudo apt-get install libboost-python-dev libboost-filesystem-dev libboost-log-dev libboost-test-dev libboost-thread-dev libboost-system-dev libgtkmm-3.0-dev libgtksourceview2.0-dev libgtksourceviewmm-3.0-dev libpython-dev libclang-dev make cmake
```
I'm working on a dependency list, so if I missed one please tell me.

eidheim:
Sure thing, after the installation later tonight.

eidheim:
Your apt-get install command was enough to compile and run juci++ on debian stable. I also manually installed g++ (doh). 

Indentation worked, as far as I tested, on my newly installed debian stable, however. Installed on VirtualBox that is.

eidheim:
Be aware though that latest debian stable (8.1) arrived 9 days ago.

PR #8, opened 2015-06-16
Code cleanup of Source::Controller and Source::Model (renamed Source::Parser)

eidheim:
First step in making the code cleaner and more C++ friendly. Hopefully I did not break anything!
See especially the comment on Source::Controller::is_saved and Source::Controller::is_changed.

### DIFF
--- a/juci/notebook.cc
+++ b/juci/notebook.cc
@@ -192,7 +192,7 @@ bool Notebook::Controller::OnKeyRelease(GdkEventKey* key) {
 
 bool Notebook::Controller::GeneratePopup(int key_id) {
   INFO("Notebook genereate popup, getting iters");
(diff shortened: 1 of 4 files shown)


eidheim:
Btw moved the source in Source::View::ApplyConfig to the Source::Controller-constructor (previously, for instance line numbers did not show in new files). This kind of initialization is ok to have in the Controller-constructor, especially when configuration options are involved (the view does not need to know about this).

zalox:
I just checked. There are no problems merging searchandreplace, but I will not merge it until it's done. There is alot of gtkmm code written by other members of the team I need to grasp in order to complete it, however it should't take longer than a couple of evenings.

The change in code style will also apply libclangmm.

PR #17, opened 2015-06-25
Cleaned up entry.*. New file now works again. 

eidheim:


eidheim:
Ready for merge if you so decide. Tried not to change especially notebook.* too much since you are also working on this (searchandreplace). Hopefully this feature is coming soon:)

Taking a break now, but its getting there. Have you looked at https://developer.gnome.org/gtkmm/stable/classGtk_1_1CssProvider.html? Maybe this can be used together with gtksourceview (GenericView) and ClangView to get a similar color profile (and for the whole application)? Leaving this up to you. Also I found this quite ok tutorial page for gtkmm: http://www.lucidarme.me/?p=4339 (css-example sadly not included).

Also fixed the terminal-window scrolling. Think I forgot to mention that. 

Some features I can look into when I got the time:
* Fallback to more seldom clang::reparse when the processing time is more than for instance 1 second (once for each get_line_count ()-change in buffer maybe). This would yield faster autocomplete, but syntax highlighting would only update when you for instance press enter. However, if we find a way to copy CXTranslationUnit this would not be needed. Also finding a way to stop CXTranslationUnit-methods would be nice. libclang has some shortcomings...
* Fix save-as
* Fix the TODO's I've added (including save dialogs on all tabs when quitting the application)
* More general cleanup
* Maybe add a singleton-class including for instance the terminal, config, and notebook classes. At the moment there are some unfortunate dependency carrying happening. Would have to be implemented with testing also in mind. And no, no dependency injection framework needed!

PR #19, opened 2015-06-30
Diagnostics on the fly

eidheim:
Had to create a better tooltip-class, the one from gtk was pretty bad. The warning and error messages are maybe a bit eager (based on mouse and text-cursor), but one can maybe add options for that in the future. 

One could also use the Tooltips class to add tooltips for parameters of class functions after completion.

### DIFF
--- a/juci/source.cc
+++ b/juci/source.cc
@@ -584,39 +584,29 @@ bool Source::ClangView::on_key_press_event(GdkEventKey* key) {
 Source::ClangViewAutocomplete::ClangViewAutocomplete(const std::string& file_path, const std::string& project_path, Terminal::Controller& terminal):
 Source::ClangView(file_path, project_path, terminal), selection_dialog(*this) {  
   get_buffer()->signal_changed().connect([this](){
+    if(last_keyval==GDK_KEY_BackSpace)
+      return;
     std::string line=" "+get_line_before_insert();
     if((std::count(line.begin(), line.end(), '\"')%2)!=1 && line.find("//")==std::string::npos) {
-      const std::regex in_specified_namespace("^(.*)(->|\\.|::)([a-zA-Z0-9_]*)$");
-      const std::regex within_namespace("^(.*)([^a-zA-Z0-9_]+)([a-zA-Z_][a-zA-Z0-9_]{2,})$");
+      const std::regex in_specified_namespace("^(.*[a-zA-Z0-9_])(->|\\.|::)([a-zA-Z0-9_]*)$");
+      const std::regex within_namespace("^(.*)([^a-zA-Z0-9_]+)([a-zA-Z0-9_]{3,})$");
       std::smatch sm;
       if(std::regex_match(line, sm, in_specified_namespace)) {
-        if(last_keyval=='.' || last_keyval=='>' || last_keyval==':') {
-          if(sm[3]=="" && !autocomplete_starting && !selection_dialog.shown) {
-            prefix="";
-            autocomplete();
-          }
-          else if(autocomplete_starting)
-            autocomplete_cancel_starting=true;
+        prefix=sm[3].str();
+        if((prefix.size()==0 || prefix[0]<'0' || prefix[0]>'9') && !autocomplete_starting && !selection_dialog.shown) {
+          autocomplete();
         }
+        else if(last_keyval=='.' && autocomplete_starting)
+          autocomplete_cancel_starting=true;
       }
       else if(std::regex_match(line, sm, within_namespace)) {
         prefix=sm[3].str();
-        if((last_keyval>='a' && last_keyval<='z') || (last_keyval>='A' && last_keyval<='Z') || (last_keyval>='0' && last_keyval<='9') || last_keyval=='_') {
-          if(!autocomplete_starting && !selection_dialog.shown) {
-            autocomplete();
-          }
-        }
-        else if(last_keyval!=0) {
-          autocomplete_cancel_starting=true;
-          if(selection_dialog.shown)
-            selection_dialog.hide();
+        if((prefix.size()==0 || prefix[0]<'0' || prefix[0]>'9') && !autocomplete_starting && !selection_dialog.shown) {
+          autocomplete();
         }
       }
-      else if(last_keyval!=0) {
+      else
         autocomplete_cancel_starting=true;
-        if(selection_dialog.shown)
-          selection_dialog.hide();
-      }
       if(autocomplete_starting || selection_dialog.shown)
         delayed_reparse_connection.disconnect();
     }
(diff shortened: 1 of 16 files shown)


eidheim:
Remember to also merge the libclang pull request before mering this.

juci/source.cc
```diff
@@ -346,6 +349,52 @@ void Source::ClangView::update_syntax(const std::vector<Source::Range> &ranges)
...
+  for(auto& diagnostic: diagnostics) {
+    if(diagnostic.path==file_path) {
+      auto start=buffer->get_iter_at_offset(diagnostic.start_location.offset);
+      auto end=buffer->get_iter_at_offset(diagnostic.end_location.offset);
+      diagnostic_tooltips.add(diagnostic.severity_spelling+": "+diagnostic.spelling, get_source_buffer()->create_mark(start), get_source_buffer()->create_mark(end));
+      auto tag=buffer->create_tag();
+      tag->property_underline()=Pango::Underline::UNDERLINE_ERROR;
+      if(diagnostic.severity<=CXDiagnostic_Warning) {
+        //TODO: get color from config.json
+        tag->set_property("underline-rgba", Gdk::RGBA("orange"));
```
zalox: I get an warning message from this, is that expected?

(juci:23904): GLib-GObject-WARNING **: g_object_set_property: object class 'GtkTextTag' has no property named 'underline-rgba'

eidheim: The underline-rgba-property came with GTK+ 3 version 3.16 (https://developer.gnome.org/gtk3/stable/GtkTextTag.html#GtkTextTag--underline-rgba). Either one could check if Gtk::TextTag got this property (I never figured out how to do this), or one could check the GTK+ 3 version before trying to set color on the underline. Will try to get the former solution to work.

eidheim: There, fixed:)

eidheim:
Also saw that one could loose focus in textview when showing tooltips on older GTK+ versions. Trying to fix this now.

zalox:
I also noticed that using CTRL+F while mouse is around the area the tooltip shows up crashes juci. The window manager I use follows the mouse with focus, meaning the tooltip had focus when trying to search. This somehow crashed juci.

zalox:
I think the same error occurs when I select a directory listing and tries to search the open document at the same time. I think a check for an active cursor could solve this.

zalox:
I also have problems selecting text with mouse.

eidheim:
The crash is happening here: https://github.com/eidheim/jucipp/blob/master/juci/notebook.cc#L239-L254. I get this warning: "Invalid text buffer iterator: either the iterator is uninitialized, or the characters/pixbufs/widgets in the buffer have been modified since the iterator was created.
You must use marks, character numbers, or line numbers to preserve a position across buffer modifications. You can apply tags and insert marks without invalidating your iterators,
but any mutation that affects 'indexable' buffer contents (contents that can be referred to by character offset) will invalidate all outstanding iterators". 

You also get this crash if you click ctrl-f, write in the sourceview, and then click next. Using marks instead of iterators fixes this, that is, keeping positions after buffer-change.

eidheim:
Indeed, I have broken selecting text with mouse! Trying to fix this now.

eidheim:
I accidentally overloaded on_motion_notify_event and on_mark_set. Fixed in latest commit.

eidheim:
I still loose focus of the textview if I keep left or right arrow pressed, and the cursor reaches a tooltip in Debian stable. This is a minor issue, might be the old GTK+ version. Textview do not loose focus if I reach a tooltip by pressing left or right arrow normally.

zalox:
If you change the tooltip only to appear on mouse hover and when using a hotkey, alot of problems will disappear. Otherwise the tooltip will flud the development process for the user.

eidheim:
We could have an option to turn it off when moving the text-cursor, but I think it should be on by default since people trying the IDE for the first time should see the warning and error messages, at least so they are aware of this feature (which no other platform independent C++ IDE has, at least not as good as this). XCode has it in some way, but its only for OS X, and it lacks c++14 support for some reason. 

A hotkey is a good idea, the user should not be forced to use the mouse while coding.

zalox:
Then it should have an intuitive way do disable it. Like the press F2 at the bottom of eclipse tooltip.

eidheim:
Sounds good!

What you show in the picture above, by the way, is the next thing I'm going to look at, that is getting information about the variables, functions and classes etc, and if that goes smoothly I guess it should not be much work to get basic renaming refactoring done (over multiple files). 

Also, if you want, I can fix the autocompletion-window, positioning etc, now that I understand all this window-buffer-root-coordinate stuff:)

zalox:
Sure!

I'm currently working on a rolling log system so that juci keeps track of the last 10 logs and makes it configurable to which log_level the logs should have.. Shouldn't take much more work, but I have a filled schedule this summer :-P

eidheim:
Got a crash when opening a .h file. Wait with the merge till I've fixed this.

eidheim:
Also got this message when opening header files: Cache global code completions. Might be able to use this to get faster code-completion for such files.

eidheim:
Seems like the crash is in libclangmm, so wait with that merge as well.

zalox:
There is possibilities in libclang to preparse files that are not open. That should be looked at as well. Imagine to "open folder" then all the files get parsed in the background and when the developer opens the file no parsing, but instant auto completion, syntax highlighting. Parsing of a big project shouldn't take longer than 1 minute, and is done only once per project.

eidheim:
Segmentation fault fixed in libclangmm, all good now.

eidheim:
The preparsing sounds interesting. One could also work with these AST files that could be produced right after running cmake, maybe to achieve the same thing.

eidheim:
Also would love to be able to copy a CXTranslationUnit, but don't seem like anyone has done this, might be a bad ida and a dead end.

zalox:
Instead of parsing the file, you can create the translationunit using this constuctor.

http://clang.llvm.org/doxygen/group__CINDEX__TRANSLATION__UNIT.html#ga0659baf7f04381286ec54b439760c8f3

eidheim:
Yes was thinking of that one. One could also maybe have a separate project-thread that created these files from time to time.

eidheim:
Tried to fix the focus loss in debian stable, but the entire application looses focus. Hopefully this is fixed in later GTK+ versions. At least the one I have on OS X is good (GTK+ 3: 3.16.3).

eidheim:
By the way,
```
namespace sigc {
  SIGC_FUNCTORS_DEDUCE_RESULT_TYPE_WITH_DECLTYPE
}
```
makes it possible to add lambdas with parameters to signals.

PR #25, opened 2015-07-15
Similar tokens now get bold property when moving insert-cursor.

eidheim:


eidheim:
There are some issues with parsing and autocomplete sometimes both being stopped after the user pauses input, typically after autocomplete running. Will fix this soon.

zalox:
I'm getting very strange behaviour with this feature. Sometimes it works as it should, I suppose, but other times it bolds randomly in the nearest area. See screenshot below.

screen

eidheim:
Can you check if the latest versions of libclangmm and jucipp works now?

zalox:
Merged this in now, it worked perfectly this time. Great work!

PR #29, opened 2015-07-17
Larger cleanup in preparation for refactoring

eidheim:
* More objects moved to singletons
* Slightly improved marking of changed tabs (works with undo at least)
* Slightly better save (not save as, haven't looked at this yet), since saving when no change will only lead to more compile time
* Refactoring moved to separate class (ClangViewRefactor).
* More use of boost::filesystem so that it works on Windows. Probably some more places that needs this. 

Also remember latest version of libclangmm (creating pull request right after this). Some more cleanup there.


eidheim:
Next thing I'll add, in a few days, is rename refactoring. But I got to cleanup my code abit first and fix some TODOs. Go to method was a nice feature though, especially when the cpp files get big! Hopefully it works well, had to struggle with the GTK's ListViewText to get it to work right.

PR #32, opened 2015-07-24
Completion improvement, fixed search and implemented replace.

eidheim:
Entry also reworked into EntryBox, which is more clean and easy to use.

Use enter to search/replace forward, and shift-enter to search/replace backward. Is the Find and Replace buttons really necessary? Guess very few use mouse clicks to search forward and backwards now. 

Also focus on entry box is fixed. For instance press tab to swap between find and replace entry-fields. 

Use escape to cancel/close entrybox.

Also, the tooltips and similar token markings are abit less eager now, only identifiers get tooltips (that is for instance (), :: -> etc no longer hold tooltips or get marked as similar tokens).  

Hmm also, should the entry box be below the sourceview? When entrybox pops up now the sourcetext is moved down. If entrybox is below the sourceview the position of the cursor will be the same on the screen.


eidheim:
Right now the syntax highlighting for clang files and non-clang files are different. I guess we can use the styles from gtksourceviewmm? With some adjustments to colors etc of course, like in the current clang-colors for light-theme.

eidheim:
See last commit. We could use these colors in json.config, for instance:
```{
    "source": {
        "colors": {
            "text_color": "text",
            "comment": "def:comment",
            ...
```
See for instance /usr/share/gtksourceview-3.0/styles/classic.xml for definition of "text" and "def:comment". We could create two such files and add them to the juci-user directory (~/.juci?). Also we could add an option in config.json to enable or disable use of dark-theme (https://developer.gnome.org/gtkmm/stable/classGtk_1_1Settings.html#a40ac901fea4b95dedf972ed5aa971c51)

eidheim:
We also need an option in config.json to select gtksourceview style.

PR #36, opened 2015-07-31
Minor improvements

eidheim:
Especially the undo/redo improvement is nice.


eidheim:
Not much left now, its getting there. I only got a few items left on my TODO:
- New menuitem under file: create C++ example project (for people testing the application for the first time mostly. This is quite essential and can thus be hardcoded into Juci, that is no need to put this into a plugin).
- Check if file/dir opened exists to avoid crash
- Use boost::filesystem everywhere (not use std::string::find("/") etc)
- Cleanup of Terminal
- Maybe make a Juci::Platform static class with various constants set depending on the operating system Juci is compiled on. The rest of the code should then be platform independent (with a few exceptions like apple-arrow keys on OS X).

eidheim:
Forgot this TODO:
- Make compile, and compile and run run in the background (using Terminal::async_execute, which I'll write in day or two).

PR #37, opened 2015-08-05
Major cleanup

eidheim:
Added cmake.h and cmake.cc to handle parsing of CMakeLists.txt files. Also fixed compile, and compile and run. Quite a few files changed, thus the pull request. 

Should find the cmake-root as well as project name in most conditions, and the executables (although, only the first one is selected in window.cc at this point). Be aware that running juci twice in Debian Stable using Gnome does not work for some reason, though other applications can be run through Juci without problem in Debian Stable.)

Remember to delete ~/.juci/config/config.json before testing. 

Going to add a way to stop the running application in the terminal at some point.


zalox:
Cool, I'm soon done with theming. It looks like we can merge without problems as I can't see many changes in the files I've been.

Syntax highlighting, gui-elements and menu all has different config files, so that's kind of messy.

eidheim:
Great, like I said in another post, it's not THAT messy with the various config-files. The new Atom editor also has at least 3 config files (1 for theme, 1 for syntax highlighting, and 1 for other stuff)

zalox:
What are your thoughts about the namespacing etc in juci now?

eidheim:
It's ok I guess, no need to add Juci:: in front of everything since this is not a library I'm thinking, but not sure.

eidheim:
By the way, with the new popen2 it is possible to stream input into the compile&run application through the terminal window. I'll fix this at some point.

eidheim:
Forgot, I added bold to stderr messages in terminal.

zalox:
I have finished the theming semantics, just need to find or create a nice looking theme. I will merge this and the themes later today if I manage :rocket:

eidheim:
Splendid!

I'm trying to fix a crash in the fork/pipe/dup/waitpid/kill-madness when running several programs from Juci. Amazing it is so difficult to start a separate process, get the streams and kill it...

zalox:
Btw, when I edit the CMakeList.txt the cmake command runs smoothly. However if I open for instance jucipp/src/CMakeLists.txt and it fails due to not being the top level CMakeLists.txt

eidheim:
There has to be a open directory with the path included in the CMakeLists.txt path for it to rerun cmake at the moment. Might be an idea to change that sure. Adding a TODO there (Notebook::save()).

eidheim:
Was hoping we could call more complex commands in Juci from a menuitem opening a command entry (like in emacs), like "make && ./src/juci", but at the moment only "./src/juci" would work. Guess its ok for now.

eidheim:
We can now run complex commands in terminal (like "make && ./src/juci")! Its soon a fully working terminal window:)

zalox:
@eidheim I'm currently working on windows support. I have managed to compile everything except Terminal. I get "invalid conversion from __FILE to int"

eidheim:
Nice! I'll look at that error Today or Tomorrow. Where in terminal.cc was the error reported?

zalox:
https://gist.github.com/zalox/aa8ab4730716a4e483aa

eidheim:
Are you compiling under Cygwin btw?

zalox:
Yes, I just installed boost, gtksourceviewmm, g++, cmake, make and git through cyg-get and compiled. I had to change some casts to compile the other classes, other than that no hiccups.

zalox:
The theming branch is updated with both your changes and mine  btw.

eidheim:
Can we merge it to the master do you think? Before the pull is getting too big? I saw one small piece missing from juci.cc (https://github.com/eidheim/jucipp/commit/ee7da8476bcfaec4552c21efa02bfd4f49c74b65), might be hard to find these later on maybe. Or is the theming not ready for the master branch just yet?

zalox:
I actually noticed it myself, I will push now.

PR #39, opened 2015-08-07
juCi++ now supports gtk-3.0 themes and gtksourceview styles

zalox:
* place themes in ~/.juci/gtk-themes/(theme_name)/gtk.css
* place gtk-styles in ~/.juci/styles/(style).xml
* select themes in config.json
```js
    "theme": {
	"main": "gtk.css",
	"theme": "dark"
    },
```


zalox:
@eidheim I made another repo with config files. If you:
```sh
$ cd && rm .juci -rf && git clone http://github.com/zalox/juci_config.git .juci
```
It should merge changes to the config folder as well.

There are some bugs I've created on purpose:
* line numbers and line highlighing does not work (pending definition from styles)

TODO: 
* Make juci clone the git repo if the config files are missing instead of using file.h
This should be easy as of the new terminal/command, or maybe there exists a git library
It could be done by downloading the zip from github as well.

Also, the theme can be changed by downloading any good old theme from http://gnome-look.org/index.php?xcontentmode=167 or something like this page. From the zip just paste the contents from a folder containing gtk-3 themes into .juci/gtk-themes/[theme]/[main] and edit the json file with the corresponding keys.

zalox:
@eidheim after I merged this there is a memory leak. I've been working with juci for an hour now and it's using 75% memory of 8GB. I haven't had time to investigate.

eidheim:
Did you work on a single file, or open/close several files, or used compile and run alot?

zalox:
The leak is somewhere in autocompletion. I tried to recreate using the global namespace, after like two or three completions the you can see the memory not decreasing after dismissing the popup.

I peeked in the implementation of clang_disposeCodeCompleteResults, it has different implementation of the destructor, maybe a hint. I don't know.

From [the source](http://llvm.org/releases/3.6.2/cfe-3.6.2.src.tar.xz) 
In file cfe-3.6.2.src/tools/libclang/CIndexCodeCompletion.cpp
```c++
void clang_disposeCodeCompleteResults(CXCodeCompleteResults *ResultsIn) {
  if (!ResultsIn)
    return;

  AllocatedCXCodeCompleteResults *Results
    = static_cast<AllocatedCXCodeCompleteResults*>(ResultsIn);
  delete Results;
}
```

eidheim:
It might also be that I create something that has to be deleted in gtkmm, like the marks one adds to the textbuffer (), I'll have a look.

zalox:
At least it narrows it down :8ball:

eidheim:
Indeed! Btw soon finished with input to the processes running in terminal. You can even run bash when selecting Project->Run Command:D Never seen an IDE before that can do that! Will also have a look at how one adds colors to bash/sh stdout, maybe that could be something cool to add too.

zalox:
Since we get such good command support I'm thinking of using git/wget to fetch configs etc instead of storing them as strings in files.h. It also simplifies the version control of the config files and branches can be used as different configurations. I couldn't get the "make install" to install the config files with the correct permissions.

eidheim:
I'm thinking that the best solution might be to have the files.h with the default file-contents. In that case, let's say that a root-user installs juci, and the user bob runs it. The creation of the .juci dir should thus not be done when installing, but rather at program start if it does not exist already.

eidheim:
Actually, the memory leak only happens on my Debian Stable, not on OS X, so it might be fixed in newer libclang versions. Will test on Debian Testing later, which hopefully have a newer libclang version.

zalox:
I use debian tesing on my laptop, where I discovered the bug...

eidheim:
Ok, will write a custom destructor then for the objects that is not freed properly in older libclang versions.

eidheim:
What are you thinking about the files.h? We keep it or?
Was thinking of adding a couple of files there for creating a new C++ project.

eidheim:
Memory leak fixed. I think I used the custom destructor because some older version of libclang in OS X did not have clang_disposeCodeCompleteResults. Not sure. Anyway, making a pull request now (libclangmm).

zalox:
@eidheim  Did you manage to change theme btw?

eidheim:
Trying to get Juci to run first:) 

I got this one when running it:
terminating with uncaught exception of type boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::property_tree::ptree_bad_path> >: No such node (clang_types)

Is files.h updated with the new nodes?

zalox:
This repo has working config files with dark theme. 

https://github.com/cppit/juci_config

If you are standing in home directory you can just

git clone https://github.com/cppit/juci_config .juci/

eidheim:
I still think we need to use files.h to populate ~/.juci. If root-user installs juci, then every user has to do this.

zalox:
I think juci should download the config files from that repo. If root user runs juci the config files will end up in root_home/.juci

eidheim:
Then the regular users have to be online to get juci to start. We need to find a better solution here. I guess the issue is the https://github.com/cppit/juci_config/tree/master/gtk-themes/dark directory. We could install this to /usr/local/share/juci maybe instead (some application-data directory on windows I guess). And the rest of the files are generated from files.h (since they are quite small).

eidheim:
Do we really need to supply this theme? Would solve alot if we only provided a nice gtksourceviewmm-style (one file which could be created from files.h), like the old one:) And let the user select different theme (from installed ones) and style in config.json. All gtk-installations also come with a dark theme, so everyone should have one they can use if they so prefer.

eidheim:
But great work! We just need to figure out the whole ~/.juci thing:)

zalox:
I made up my mind:
* we supply information about how to add themes
* we implement the style selector (bundled with gtksourceview) and merge it with a dropdown with selection of gtk-themes (should be a simple inherit thingy)
* continue using files.h

eidheim:
Sounds good!

zalox:
I tried to compile with clang instead and it complains about "thread local storage."

zalox:
´´´sh
$ make
Scanning dependencies of target juci
[  6%] Building CXX object src/CMakeFiles/juci.dir/juci.cc.o
In file included from /home/zalox/juci/src/juci.cc:1:
In file included from /home/zalox/juci/src/juci.h:4:
In file included from /home/zalox/juci/src/window.h:7:
In file included from /home/zalox/juci/src/notebook.h:6:
In file included from /home/zalox/juci/src/source.h:9:
/usr/lib/gcc/x86_64-pc-cygwin/4.9.3/include/c++/mutex:707:10: error: thread-local storage is not supported for the current target
  extern __thread void* __once_callable;
         ^
/usr/lib/gcc/x86_64-pc-cygwin/4.9.3/include/c++/mutex:708:10: error: thread-local storage is not supported for the current target
  extern __thread void (*__once_call)();
         ^
2 errors generated.
src/CMakeFiles/juci.dir/build.make:54: recipe for target 'src/CMakeFiles/juci.dir/juci.cc.o' failed
make[2]: *** [src/CMakeFiles/juci.dir/juci.cc.o] Error 1
CMakeFiles/Makefile2:77: recipe for target 'src/CMakeFiles/juci.dir/all' failed
make[1]: *** [src/CMakeFiles/juci.dir/all] Error 2
Makefile:117: recipe for target 'all' failed
make: *** [all] Error 2
´´´

eidheim:
I guess gcc is the way to go on Cygwin since its most tested and most libraries there are compiled with GCC already. You might get strange errors when trying to link a clang-program to a gcc library I think. Also those error do not make sense to me:)

zalox:
Yeah, the last one didn't make sense at all. Maybe the pipe() and close() methods have different return types, or that the compiler is confused about some deductions.

eidheim:
By the way, can I clean up the theme/style thing? There will be only two options (hopefully) in config.json, theme and style. We can start out with the default themes and styles, with an option to add more into ~/.juci/style and ~/.juci/theme directories if the user would like that. Maybe we could provide a nice style. Also, dark theme does sadly not work well on OS X, since popups and menuitems start out white, before they become dark. Might be best therefore, to have the default style/theme light if it is ok?

Will also look at the clang-codes and see if I can find something that works on most gtksourceview-styles (including the tooltip-background). Not all have the "def:..." ones.

zalox:
Sure, go ahead! I will focus on windows support! I'm thinking of supplying a working download for W8 and W10 as a portable install.

PR #41, opened 2015-08-13
Theme cleanup

eidheim:
If we go for a light theme as default, since dark GTK-themes are pretty bad on OS X, should we go for your old nice looking style:)? One could call it juci and add it under ~/.juci/styles (which is already created and is in the style search path). I could create one quick Tomorrow, based on your old config.json file, after adding a few extra tags for tooltips and warning and the like, and you can adjust later on when you feel like. Then people would see which tags Juci is searching for as well in clang-mode. 

Also, one option, is to go for a default light theme/style on OS X, and a dark default on Linux and maybe Windows if they are working ok there. This is easy to do in files.h (#ifdef __APPLE__ and so on). But I guess this could be up to the user. Changing themes now is very easy (if one knows what they are called!), and there are loads installed already in a default gtk-install. All the themes seem also to have two variants; one dark and one light.


eidheim:
Ah, forgot, remember to delete the ~/.juci-directory to get the latest changes.

eidheim:
Juci now handles æøå correctly (for instance in comments, strings and the like!

eidheim:
I forgot, also get the latest libclangmm before compiling.

eidheim:
The monospace font option in sourceview is important, since text will move around like crazy if bold for instance is toggled. The rest of the application uses the theme default font.

eidheim:
I'll leave the styles for now, it's working ok, and we got at least a dark style that does not melt your eyes (to be used together with dark themes).

PR #45, opened 2015-08-30
A quick pull request of two neat commits till Monday

eidheim:
If ok, could we merge these two commits before Monday morning?

### DIFF
--- a/src/entrybox.cc
+++ b/src/entrybox.cc
@@ -4,12 +4,43 @@ namespace sigc {
   SIGC_FUNCTORS_DEDUCE_RESULT_TYPE_WITH_DECLTYPE
 }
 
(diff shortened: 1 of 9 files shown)


eidheim:
Tested and works on Debian stable, OS X and Windows.

zalox:
I will also introduce the "about popup" before tomorrow morning.

eidheim:
Splendid!

eidheim:
The other commits seem to be working as well on all the OS'es I have.

zalox:
If I get time I will try to make a bot to compile the master each time it updates. At least on debian and ubuntu.

zalox:
I just merged this. I introduced version number 0.9.1 for some reason.

PR #54, opened 2015-09-08
Many bug fixes and improved clang indentation

eidheim:
Tested on Ubuntu 14, debian stable/testing, OS X and MSYS2. 

The clang indentation should now work flawlessly (hopefully) in every case, even in multi-line expressions. 

Would be nice if we could merge this soon, after you have tested it of course. If there are issues, I will try fix it Today or Tomorrow, but I will not commit more features before it's merged:)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,7 +3,9 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O3 -std=c++11 -pthread -Wall -Wno-reord
 set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_LIST_DIR}/cmake/Modules/")
 
 if(APPLE)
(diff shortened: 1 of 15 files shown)


zalox:
Can you try to open api.cc and wait until parsing is done? It crashes on debian testing sometimes.

zalox:
I'm using libboost1.55-all-dev btw, since debian testing has 1.58 as well.

eidheim:
libclang: crash detected during reparsing

Creating an issue on this, it also crashed on OS X

eidheim:
We'll fix this at a later point I guess.

eidheim:
Is the indentation acceptable do you think? Later I'll add "set indentation char and size" and "auto indent file" to the menu.

zalox:
This is the first time indenting has worked on my laptop for some reason. So I like it.

eidheim:
I know there was a problem with key states in Ubuntu, but that is fixed now.

eidheim:
Btw, we should maybe read the menu-xml directly from files.h instead of .juci/... Maybe that will fix the Windows menu issue as well?

PR #57, opened 2015-09-13
Using windows dialogs instead of gtk + cleanup of dialogs.

zalox:
Still buggy/non-working on windows. I will continue later this week.


zalox:
Finally after some struggle I made some progress. I compiled a "msvs .dll" and used some cool tricks to make it link with the gcc linker. (So it works in msys2) I guess this approach may be useful later. 

The "msvs .dll" can be precompiled for 32-bit and 64-bit and released on this repo.

Install size of juci after this approach is 32 mb, zipping it will reduce this to about 15 mb.

zalox:
@eidheim This is ready to be merged. It works as expected when run from msys2, if extracting the juci.exe and dlls there are still some problems with paths.

zalox:
Also if you don't mind. Take a look at https://github.com/cppit/dialogs/blob/master/dialogs_win.cpp for the implementation of the dlls.

eidheim:
Splendid! I'll have a look Tomorrow morning.

zalox:
Btw, the dlls were compiled in W8.1 using the latest version of MSVS Community Edition. You can compile the gitrepo by "File > New Project > From existing code". By opening the project this way we get rid of all the msvs generated hpp- and cpp-files. You may also need to edit some properties for generation of dlls instead of a static lib.

eidheim:
Why the dll's? One can't just compile the dialog source into juci?

zalox:
No, the windows headers are just available for msvc.

eidheim:
Aha!

Are there any license implications of using a library compiled with MSVC?

src/singletons.h
```diff
@@ -11,6 +11,12 @@
 #include <gtkmm.h>
 #include <string>
 
+#ifdef _WIN32
+#define HOME "AppData"
```
eidheim: Call the variable JUCI_HOME, there might be a variable already called HOME on some system. Also, what if running juci in MSYS2? Is the home directory then not stored in the HOME env? Might be better to make a static function (get_home_path() or something in juci::filesystem) that checks if getenv("HOME") returns something, if not, use getenv("AppData"), on Windows that is.

zalox:
"An unlimited number of users within an organization can use Visual Studio Community for the following scenarios: in a classroom learning environment, for academic research, or for contributing to open source projects."

The api is proprietary, but free to use for open source and individuals.

eidheim:
One could add a comment in https://github.com/zalox/jucipp/blob/master/WIN-packages/dialogs_win.h linking to the source of the .dll files (https://github.com/cppit/dialogs). Compiled libraries are usually ok under most licenses, as I understand it, as long as people can compile them themselves somehow.

src/window.cc
```diff
@@ -29,8 +30,8 @@ void Window::generate_keybindings() {
   }
 }
 
-Window::Window() : box(Gtk::ORIENTATION_VERTICAL), notebook(directories), compiling(false) {
-  DEBUG("start");
+Window::Window() : box(Gtk::ORIENTATION_VERTICAL), notebook(*Singleton::directories()), compiling(false) {
```
eidheim: No need to pass directories to notebook anymore, since directories are now a Singleton

src/window.cc
```diff
@@ -165,23 +166,94 @@ void Window::create_menu() {
     hide();
   });
   menu.action_group->add(Gtk::Action::create("FileNewFile", "New File"), Gtk::AccelKey(menu.key_map["new_file"]), [this]() {
-    new_file_dialog();
+      boost::filesystem::path path = Dialog::new_file();
```
eidheim: Minor: indentation is wrong (4 spaces instead of 2)

src/window.cc
```diff
@@ -165,23 +166,94 @@ void Window::create_menu() {
...
+        else {
+          if(juci::filesystem::write(path)) {
+            if(Singleton::directories()->current_path!="")
+              Singleton::directories()->update();
+            notebook.open(path.string());
+            Singleton::terminal()->print("New file "+path.string()+" created.\n");
+          }
+          else
+            Singleton::terminal()->print("Error: could not create new file "+path.string()+".\n");
+        }
```
eidheim: Minor: indentation is wrong (4 spaces instead of 2)

src/window.cc
```diff
@@ -165,23 +166,94 @@ void Window::create_menu() {
...
+        if(boost::filesystem::last_write_time(path)>=time_now) {
+          if(Singleton::directories()->current_path!="")
+            Singleton::directories()->update();
+          Singleton::terminal()->print("New folder "+path.string()+" created.\n");
+        }
+        else
+          Singleton::terminal()->print("Error: "+path.string()+" already exists.\n");
+      } else {
+        Singleton::terminal()->print("Cancel \n");
+      }
```
eidheim: Minor: indentation is wrong (4 spaces instead of 2)

src/window.cc
```diff
@@ -165,23 +166,94 @@ void Window::create_menu() {
...
+        Singleton::terminal()->print("Error: "+cpp_main_path.string()+" already exists.\n");
+        return;
+      }
+      std::string cmakelists="cmake_minimum_required(VERSION 2.8)\n\nproject("+project_name+")\n\nset(CMAKE_CXX_FLAGS \"${CMAKE_CXX_FLAGS} -std=c++1y -Wall\")\n\nadd_executable("+project_name+" main.cpp)\n";
+      std::string cpp_main="#include <iostream>\n\nusing namespace std;\n\nint main() {\n  cout << \"Hello World!\" << endl;\n\n  return 0;\n}\n";
+      if(juci::filesystem::write(cmakelists_path, cmakelists) && juci::filesystem::write(cpp_main_path, cpp_main)) {
+        Singleton::directories()->open(project_path);
+        notebook.open(cpp_main_path);
+        Singleton::terminal()->print("C++ project "+project_name+" created.\n");
+      }
```
eidheim: Minor: indentation is wrong (4 spaces instead of 2)

src/window.cc
```diff
@@ -165,23 +166,94 @@ void Window::create_menu() {
...
+      else
+        Singleton::terminal()->print("Error: Could not create project "+project_path.string()+"\n");
   });
   menu.action_group->add(Gtk::Action::create("FileOpenFile", "Open File"), Gtk::AccelKey(menu.key_map["open_file"]), [this]() {
-    open_file_dialog();
+      notebook.open(Dialog::select_file());
```
eidheim: Minor: indentation is wrong (4 spaces instead of 2)

src/window.cc
```diff
@@ -165,23 +166,94 @@ void Window::create_menu() {
...
   });
   menu.action_group->add(Gtk::Action::create("FileOpenFolder", "Open Folder"), Gtk::AccelKey(menu.key_map["open_folder"]), [this]() {
-    open_folder_dialog();
+    auto path = Dialog::select_folder();
+    if (boost::filesystem::exists(path))
+      Singleton::directories()->open(path);
+    else
+      Singleton::terminal()->print("Cancel \n");
```
eidheim: Use a more appropriate error message like "Error: directory does not exist"

src/window.cc
```diff
@@ -165,23 +166,94 @@ void Window::create_menu() {
...
+          file << notebook.get_current_view()->get_buffer()->get_text();
+          file.close();
+          if(Singleton::directories()->current_path!="")
+            Singleton::directories()->update();
+          notebook.open(path);
+          Singleton::terminal()->print("File saved to: " + notebook.get_current_view()->file_path.string()+"\n");
+        }
+        else
+          Singleton::terminal()->print("Error saving file\n");
+      }
```
eidheim: Minor: indentation is wrong (4 spaces instead of 2)

src/window.cc
```diff
@@ -165,23 +166,94 @@ void Window::create_menu() {
...
+        if(file) {
+          file << notebook.get_current_view()->get_buffer()->get_text();
+          file.close();
+          if(Singleton::directories()->current_path!="")
+            Singleton::directories()->update();
+          notebook.open(path);
+          Singleton::terminal()->print("File saved to: " + notebook.get_current_view()->file_path.string()+"\n");
+        }
+        else
+          Singleton::terminal()->print("Error saving file\n");
```
eidheim: "Error: could not save "+path+"\n"

src/window.h
```diff
@@ -12,7 +12,6 @@
 class Window : public Gtk::Window {
 public:
   Window();
-  Directories directories;
```
eidheim: Also remove 
Directories &directories;
in directories.h

eidheim:
After fixing the comments I think this looks good. Will test after they are resolved. 

Are there a reason behind the C functions in https://github.com/cppit/dialogs/blob/master/dialogs_win.cpp?
For instance:
const char* c_select_file() {
  return (OpenDialog("Open file", 0)).show().c_str();
}
can't be:
std::string dialogs_win_select_file() {
  return (OpenDialog("Open file", 0)).show();
}
? (that is without the extern "C" I guess then)
It does not matter much, but at least one copy is avoided.

src/window.cc
zalox: I actually changed this, but I lost it when merging with master.

src/window.cc
zalox: I will expand the functionality of the dlls to discover cancel, X, esc etc. This is basically an quickfix.

zalox:
This is due to the visual compiler. When linking the headers gcc can't understand c++-dlls, since msvc translates method signatures differently. msvc, clang, gcc, intel all translate the same way in C, making linking possible.

The extern C makes sure the naming happens in C-style, but only allowing C types.

zalox:
Ready for merge I think.

eidheim:
Great, I'll test it Tomorrow

eidheim:
Merging this soon, after having made some smaller changes. There are some issues still, but I'll fix them after I merge #68 as well.

PR #64, opened 2015-10-02
Quick pull request to fix Ubuntu 14 compilation

eidheim:
Also a speed up when using source map with huge source files.


eidheim:
@zalox I'll merge this right away so that jucipp compiles on Ubuntu 14.

PR #73, opened 2015-10-17
Minor fixes + home directory now validated.

zalox:


eidheim:
By the way, nevermind terminal messages on cancel in the dialogs. We do not want to flood the terminal with such messages, since the user most likely know that he/she canceled the operation.

zalox:
Okei, I'll add it as a log entry instead.

eidheim:
Or if we want to exit the application in case of no home path write an appropriate error message to std::cerr and exit somehow.

eidheim:
Btw, there are a few things I'd like to do before we create release v0.9.3 of juci this weekend:
* fix dialogs slightly, mostly the save dialog, and attach a window to the dialogs in gtk, since an error message gets written to std:cerr
* Finish the auto-indent buffer menu option using astyle (external program that hopefully is installable on all platforms). 
* Cleanup terminal*

zalox:
I have an solution for dialogs, I'm also making this an submodule for both windows and gtk.

eidheim:
More submodules? Lets try to keep that to a minimum please. Its fine the way it is now in my opinion.

zalox:
I fixed the error to std::err and removed exceptions as much as possible.

Can't you indent everything using clang?

eidheim:
For auto-indenting a complete file/buffer:
Seems like clang-format is the only executable that is easily installable on MSYS2, and the rest of the platforms, however, both uncrustify and astyle supports some more languages. 
We can go with clang-format for now.

eidheim:
Is it possible to make the Windows dialogs open the directory path as start-path as when using the gtk dialogs? Also the Windows dialog should open the current MSYS2 path when run from MSYS2 if there is no directory opened. 

Also, the libdialogs.dll is not installed in MSYS2. MSYS2 version is thus not currently working unless one does this manually.

eidheim:
Good set_transient_for fix!

Additionally, the gtk dialog save as should have the current file path as proposal, and show the content of the current file path directory (as before). I leave this up to you:) You must most likely add a const std::string& file_path argument to this function.

zalox:
Wrapping all singletons into Singletons is essential for the plugin api. I would also like to shorten the footprint of Singleton::window(), Singleton::terminal()->print() etc. api::println();

The "standing" directory problem will be fixed soon. I will look at it tonight or tomorrow.

eidheim:
If you want to look at the python bindings, there might be a chance that gtk has some way of doing this, that is, instead of using boost::python. Its worth looking into, as then maybe one could use the gtk python bindings as well in the python plugins.

In my opinion the use of singleton objects and their functions is at the moment describing and intuitive. Not so sure about api::println at least in the c++ code.

eidheim:
I managed to compile this in MSYS2:
dialogs_win.cpp:
```c++
#define NTDDI_VERSION NTDDI_VISTA
#define _WIN32_WINNT _WIN32_WINNT_VISTA

#include "dialogs.h"
#include <windows.h>
#include <shobjidl.h>

std::string Dialog::select_folder() {
  return "";
}

std::string Dialog::new_file() {
  return "";
}

std::string Dialog::new_folder() {
  return "";
}

std::string Dialog::select_file() {
  HRESULT hr = CoInitializeEx(NULL, COINIT_APARTMENTTHREADED | COINIT_DISABLE_OLE1DDE);
  if (SUCCEEDED(hr)) {
    IFileOpenDialog *pFileOpen;
    hr = CoCreateInstance(CLSID_FileOpenDialog, nullptr, CLSCTX_ALL, IID_IFileOpenDialog, reinterpret_cast<void**>(&pFileOpen));
    if (SUCCEEDED(hr)) {
      hr = pFileOpen->Show(nullptr);
    }
  }
  
  return "";
}

std::string Dialog::save_file() {
  return "";
}
```
The trick was the defines, and most likely we can remove them later when MSYS2 gets better Windows support

Thus we can get rid of the dll's and instead write dialogs_win.cpp similar to write dialogs.cpp for gtk. This will make the dialog source more clean AND we get rid of the precompiled dll's. 

I leave it up to you to rewrite src/dialogs_win.cpp.

eidheim:
I got this, by the way, from https://github.com/rafalcieslak/nativefiledialog/commit/2ec8d20ce8fe15564e7ad0f6a3502e7acf6845ab

https://github.com/rafalcieslak/nativefiledialog is similar to what you have done, but more complex. I like your solution much better:)

eidheim:
Great work, we should try merge soon, before we get double digit conflicts:) Let me know when you are done with the most important commits towards creating our first stable release, and I'll merge and resolve the two forks. Have time Wednesday and Saturday for this.

eidheim:
Is this ready for merge?

zalox:
I have some more commits comming up, reflecting some of your comments. Stand by!  :+1:

zalox:
Tested on windows and debian testing. Ready for merge.

eidheim:
Great! By the way, we need an juCi++ icon! Have to create a juci.desktop file for Ubuntu or else "Unknown application name" will show in the menu (the final Ubuntu menu bug that needs to be fixed), and in this file one can include an icon:)

eidheim:
Merging the two pull requests now, and fixing conflicts.

eidheim:
Fixing things so that one can use `Singleton::terminal()->print("Error reading "+config_json + "config.json: "+e.what()+"\n");` in config.cc, or else it will be really hard to figure out what is wrong with the config.json one just edited.

eidheim:
Also fixed it so the config files now gets read again after saving config.json

eidheim:
I think maybe we should initialise the lazy singletons in app::app() (juci.cc), abit chaotic atm when one does not exactly know when they get initialised. I'll fix it.

eidheim:
My TODO before we got a stable release:
* speed up smart paste if possible
* create file on `juci filename.txt` if it does not exist (most of the cases you want to create a file so I guess that's ok)
* Fix "Unknown Application Name" on Ubuntu. We then need to install a juci.desktop file through cmake along with a cool looking icon (I think you're better at creating this icon than me, maybe "uC++" or something with colors?)
* Doublecheck if smart-keys works if one combines tabs and spaces
* various bug fixes and cleanups (I'll fix terminal.* later, seems like people need a platform-independent process library, I'll try to do that next week). 
* See if there is an option in clang-format to keep empty line spaces, and use this if some option is on to keep these when saving.

PR #75, opened 2015-10-26
Major change: replaced old soon to be deprecated menu system, with the new menu system

eidheim:
* The new menu system uses native menu functions (looks especially good on OS X)
* Keybindings gets updated if changed in config.json (no need to restart juci anymore)
* Auto-indentation using clang-format (do NOT use on juci source files since it lead to crazy commits)
* Fixed a crash
* Added nice looking background pattern in the styles (for gtksourceview versions 3.16 and above)
* Some minor fixes and cleanups

Not yet fully tested on all platforms

Edit: added info on keybindings


eidheim:
Not working yet on Ubuntu 14 due to buggy gtk/gtkmm 3.10 package.

eidheim:
There are issues on Ubuntu with the new menu system (they sadly have their own buggy menu system):
* Ubuntu cant find application name
* Changed shortcuts does not show in the menu

Will try to disable the buggy Ubuntu menu system, as it is not even fixed in Ubuntu 15.....

zalox:
I think we can protect this project by making git take care of line endings, unicode, ascii internaly, making us immune against clang-format issues.

https://help.github.com/articles/dealing-with-line-endings/

eidheim:
crlf/lf should not be an issue as long as you write in juci only, it always save newlines as lf. However, the main issue is keeping/not keeping empty line spaces like "     ". Atom for instance has default option to replace empty line spaces "    " with "" when saving, and this may lead to ugly commits. Clang-format also does this, not sure if there is an option yet to disable this. Juci currently removes empty spaces after expressions when you press enter, but replacing empty line spaces with empty lines when saving feels too disrupting to repositories that has alot of empty line spaces. 

Not entirely sure what to do with respect to this actually, should maybe have an option to remove empty line spaces in config.json. Not sure though if it should be enabled by default.

PR #76, opened 2015-10-28
More configuration

zalox:
* clang-format can now be edited in config.json 
* directories automatically update when editing filters

Please review and merge

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -169,9 +169,6 @@ void MainConfig::GenerateDirectoryFilter() {
     dir_cfg->exceptions.emplace_back(i.second.get_value<std::string>());
   for ( auto &i : ignore_json )
     dir_cfg->ignored.emplace_back(i.second.get_value<std::string>());
-  if(Singleton::directories() != nullptr) {
-    Singleton::directories()->update();
-  }
 }
 
 void MainConfig::init_home_path(){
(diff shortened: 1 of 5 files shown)


src/config.cc
```diff
@@ -71,6 +71,7 @@ void MainConfig::retrieve_config() {
   Singleton::Config::window()->default_size = {cfg.get<int>("default_window_size.width"), cfg.get<int>("default_window_size.height")};
   Singleton::Config::terminal()->make_command=cfg.get<std::string>("project.make_command");
   Singleton::Config::terminal()->cmake_command=cfg.get<std::string>("project.cmake_command");
+  Singleton::Config::terminal()->clang_format_command=cfg.get<std::string>("project.clang_format_command", "clang-format");
   Singleton::Config::terminal()->history_size=cfg.get<int>("terminal_history_size");
```
zalox: @eidheim Maybe we should add default values for all the property_tree get methods so they become exception safe?

eidheim: Have thought about it, the problem is one has to write the default values twice, and that is not a good programming practice.

zalox: I have an idea solving this without duplication, i'll make a pull request ~tomorrow with a suggestion.

eidheim: Great! Also try to prioritize the remaining bugs, like the Windows dialogs if they are not yet working like the gtk-ones. I noticed for instance that one had to create a directory first instead of just writing the name of a new project when selecting New Project->C++ on Windows. Also, when running juci++ from msys-terminal, the home directory is not /home/[username], but /Users/... Maybe this is intended, not sure myself what is best here.

PR #80, opened 2015-11-02
Singleton and config cleanup, also some minor fixes

eidheim:
Not yet tested on all platforms. Keep in mind that Window is no longer a singleton. This was done to remove all the circular includes. Window can still be reached, look at dialogs.cc for an example. 

Compilation time also greatly improved.


eidheim:
@zalox: There are some issues with the dialogs_win.cc that I am currently fixing. Should be done Tomorrow morning if not before. I know you are working on the same, but the reason behind it was segmentation faults. Hopefully the work is not overlapping too much.

zalox:
I'll make an merge request now. I changed a lot.

eidheim:
Splendid, I'll continue on it Tomorrow after applying your changes.

eidheim:
I went maybe abit crazy on dialogs_win.cc, but there were some issues. Mainly:
* Exceptions are not handled correctly always on MSYS2 it seems, may lead to segmentation faults
* Save file as extensions were set only to cpp-files, now it detects the extension used on the file that is going to be saved.
* No need for 3 classes when one can do with 1
* Opened folder is now the same as the gtk version
* Memory was not always freed

I must say, I feel sorry for the people working on the Windows API every day... Hopefully, I will never see that awful thing again!;)

zalox:
Please notify when ready for merge, I've looked over the changes.

eidheim:
Tested on most platforms. Ready for merge I think.

PR #81, opened 2015-11-02
Dialogs now has equal behaviour

zalox:
* Introduce boost.locale dependency in Msys, no need to update dependencies since it is automatically installed when installing boost
* Fix crashes due to special characters
* Correct start directory is now selected, slightly different behavior than gtk due to windows conventions
* Default file extension is selected as .cpp 
* Less copies due to move semantics

### DIFF
--- a/src/dialogs_win.cc
+++ b/src/dialogs_win.cc
@@ -1,7 +1,7 @@
 #ifdef _WIN32
 #include "dialogs.h"
 #include "singletons.h"
-
+#include <boost/locale.hpp>
 #ifndef check
 HRESULT __hr__;
 #define check(__fun__, error_message)            \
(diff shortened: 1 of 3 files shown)


zalox:
Please merge!

src/dialogs_win.cc
```diff
@@ -12,68 +12,66 @@ HRESULT __hr__;
...
 CommonDialog::CommonDialog(CLSID type) : dialog(nullptr) {
     check(CoCreateInstance(type, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&dialog)), "Failed to create instance");
     check(dialog->GetOptions(&options), "Failed to get options from instance");
 }
-
-void CommonDialog::set_title(const std::string &title) {
-    auto tmp = std::wstring(title.begin(), title.end());
-    auto t = tmp.data();
-    check(dialog->SetTitle(t), "Failed to set dialog title");
+void CommonDialog::set_title(std::string &&title) {
```
eidheim: In general, that is, 99.9% of the cases or so, we do not need to implement or use move operator specifically unless we for instance write our own container that we use. I'll fix this in #80.

PR #85, opened 2015-11-08
Mostly minor cleanup and fixes, added two new features.

eidheim:
New:
* can do `juci . new_file.py` for instance
* added option to remove trailing whitespace characters and add trailing newline if missing on save. Off by default since it create annoying commits if a git-repository contains trailing whitespace characters already.


eidheim:
Tested and works on OS X, MSYS2, Debian stable/testing and Ubuntu 14

PR #86, opened 2015-11-14
Desktop file for Linux

eidheim:
Merging right away since it fixes mostly the Ubuntu menu bug. Also one can now open files through a file manager using juCi++.

### DIFF
--- a/share/juci.desktop
+++ b/share/juci.desktop
@@ -1,6 +1,5 @@
 [Desktop Entry]
-Name=juci
-GenericName=juCi++
+Name=juCi++
 Comment=A lightweight IDE
 Exec=juci %U
 Terminal=false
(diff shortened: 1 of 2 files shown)


share/juci.desktop
```diff
@@ -0,0 +1,8 @@
+[Desktop Entry]
+Name=juCi++
+Comment=A lightweight IDE
+Exec=juci %F
+Terminal=false
+Type=Application
+StartupNotify=true
+MimeType=text/plain;
```
zalox: We should add an category as well. Programming for instance.

eidheim: Is this ok:
```
Categories=Development;GTK;
```
I see Development is commonly used for applications related to programming. GTK since its main GUI library is GTK.

PR #87, opened 2015-11-15
Go to Usage implementation

eidheim:
Note that like rename refactor, Go to Usage only considers open and parsed buffers.


eidheim:
Added support for more platforms through using boost::regex instead of std::regex, and simplified installation on Ubuntu 14 and Mint. Should be pretty ready for merge I think.

PR #91, opened 2015-11-24
Little fixes

VadimGit:


eidheim:
Saw you restored boost::regex. The reason behind boost::regex is that older systems like Ubuntu 14 does not support std::regex since they are still using gcc 4.8.

VadimGit:
>Have to override hide to be able to ask if one wants to save any unsaved buffers when quitting. 

from file /usr/include/gtkmm-3.0/gtkmm/widget.h
````cpp
public:
  //C++ methods used to invoke GTK+ virtual functions:
protected:
  //GTK+ Virtual Functions (override these to change behaviour):
  //Default Signal Handlers::
  // This is a default handler for the signal signal_show().
  virtual void on_show();
  // This is a default handler for the signal signal_hide().
  virtual void on_hide();
````

eidheim:
You cannot cancel hide in on_hide, that is why hide is overridden instead. As long as you call hide from a dynamic casted (juci::)Window object it is ok. Gtkmm is not perfect, but it is the best alternative for modern C++. 

You can test this by opening juci, changing a file without saving it, quit juci, then click close window (not yes or no) on the dialog that comes up.

VadimGit:
I can't test it. I don't see a difference. with hide() and on_hide() windows behave same. Do you try to disable hiding in this way?

eidheim:
Yes, if the save changed buffer dialog is closed, instead of answering yes or no, juci does not quit. This is common for most if not all editors.

VadimGit:
behave like "there are unsaved changes Do you really want to quit?" if "no' return to editor?

eidheim:
When you press esc-key for instance instead of answering yes or no in the dialog.

eidheim:
Will merge this later Today or Tomorrow when I have time, will also fix all the new override warnings that came when introducing override in juci.h.

PR #93, opened 2015-11-26
Fixes #92 and use of ' in spellchecks, and some cleanups.

### LINKED ISSUE #92: Changing CMakeLists.txt when having many files open
If I have many files open my computer melts because of async reparse of all the files at once.

Options:
1. Queue the reparse
2. Reparse only visible file, and others when changed to

eidheim:


eidheim:
Tested on most platforms. Read for merge I think.

src/source_clang.cc
```diff
@@ -874,8 +874,8 @@ void Source::ClangViewAutocomplete::async_delete() {
   parse_thread_stop=true;
   delete_thread=std::thread([this](){
     //TODO: Is it possible to stop the clang-process in progress?
-    if(restart_parse_thread.joinable())
-      restart_parse_thread.join();
+    if(full_reparse_thread.joinable())
```
zalox: Shouldn't it be possible to terminate the thread or something? Kill the pid or something like that?

eidheim: Sadly, there is no way to stop a clang-parse/reparse or autocomplete once it has been started (at least I have not found it yet). Normally, this would be done by for instance setting some stop_process atomic variable to true, but libclang has no such method.

eidheim: Terminate a thread is impossible to perform without leaving objects that have not been freed, as I understand it. Thus this is never part of a thread-library. Processes, on the other hand, can be killed, and their memory is freed by the OS.

PR #94, opened 2015-11-27
Update submodule

zalox:


eidheim:
Ah yeah, also I have to create a pull request using the new libclangmm methods (I have committed the code to eidheim/jucipp before we do this merge.

eidheim:
See #95.

eidheim:
Maybe the absolutely best way to do it is to create the link in #95 so that the new link and the code utilising the new libclangmm methods are merged together.

eidheim:
Closing this pull request, I'll add the link in #95, and it will then be easier to test #95 for us both when it is ready Tomorrow or so.

PR #96, opened 2015-11-30
Little fixes

VadimGit:


eidheim:
I agree, this looks better. Will merge Tomorrow after testing that gcc 4.8 on Ubuntu 14 handles the emplace_back template.

PR #98, opened 2015-12-04
Shiny new process class and cleanup of terminal*

eidheim:
* Can now do for instance "echo test && echo test" on Windows, as the Process class uses sh if compiled with -DMSYS_PROCESS_USE_SH. 
* Now attempts to kill child processes of a started process on Windows if any, as far as Windows actually manage to do this. 

Tested on MSYS2, OS X and Ubuntu 14.


eidheim:
Forgot: ready for merge if your tests go through.

eidheim:
Tested again on OS X, MSYS, Ubuntu 14. And also Debian testing.

src/terminal.cc
```diff
@@ -0,0 +1,275 @@
...
+void Terminal::InProgress::cancel(const std::string& msg) {
+  if(!stop) {
+    stop=true;
+    Singleton::terminal->async_print(line_nr-1, msg);
+  }
+}
+
+Terminal::Terminal() {
+  bold_tag=get_buffer()->create_tag();
+  bold_tag->property_weight()=PANGO_WEIGHT_BOLD;
```
zalox: Why not use enum directly? In terminal->print() we could add PANGO_WEIGHT_BOLD | PANGO_STYLE_ITALIC for instance.

zalox: Saw that this is texttag which doesn't have operator|() overloaded so I guess just ignore this.

eidheim:
@zalox : If this looks ok, you can merge and pull into your PR (have tested it thoroughly on all platforms). In that way we only have to resolve a few  conflicts at a time, and it will be easier to find errors in the conflict resolution, if any.

PR #108, opened 2015-12-11
Add screenshots

zalox:
Update screenshots to be on top of description

- [x] make sure links points to master
- [x] ready for merge


zalox:
@eidheim I think we should have a banner like screenshot like I've added here.

eidheim:
Some issues:
* Try not use the emacs dark blue in the screenshots, someone might feel we steal their colors
* Avoid white theme with dark style, this is a remnant from ide's that did not manage to change both
* The minimap takes up way too much space, people will definitely comment this
* You have 1 error (in bottom left), even though no error should show (probably because you deleted the error code and did a autocomplete after without a reparse). This also looks strange. 
* A better c++14 example I think may be:
```c++
#include <chrono>
...
auto time=10s;

cout << time.
```
* Some does not even include a screenshot, like https://github.com/atom/atom
* If they do include a screenshot, the best place would maybe be below features, and smaller clickable screenshots that do not take up too much space (like https://github.com/hioa-cs/IncludeOS)

Sorry for the rant like item list:)

eidheim:
I'm merging what we both agree upon now, that is the screen shots below features. I should have created a separate branch for this, will do that next time, so by merging now I do not create weirdness with my master branch.

eidheim:
@zalox feel free to create another screenshot branch if you want to change something, and we can discuss it there. Hopefully we can both come to an agreement in the end:)

PR #116, opened 2015-12-14
Announce

### LINKED ISSUE #103: New file should open empty "untitled" file not save dialog
I think the behaviour is to aggressive with an popup.

An empty buffer would be better in my opinion.

zalox:
I'd like to discuss a few adjustments before release.

## 1. Open config.json by default on startup
 - This gives the developer intimidate access to configuration which we all love. 
 - We should also add a greeting to the user as an option

By the greeting I mean an option containing a greeting :palm_tree: 

## 2. Close #103 
- by supporting eidheims arguments :)

## 3. Add docs/how-to-contribute.md
- Explain what coding standards, naming and line endings etc. we expect in PRs

edit: formating

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -5,4 +5,4 @@
 [submodule "tiny-process-library"]
 	path = tiny-process-library
 	url = https://github.com/eidheim/tiny-process-library
-	branch = v1.0.1
+	branch = v1.0.2
--- a/docs/install.md
+++ b/docs/install.md
@@ -31,6 +31,23 @@ make
 sudo make install
 ```
 
+##Arch Linux
+Install dependencies:
+```sh
+#as root
+pacman -S git cmake make clang gtksourceviewmm boost aspell aspell-en
+```
+
+Get juCi++ source, compile and install:
+```sh
+git clone --recursive https://github.com/cppit/jucipp
+cd jucipp
+cmake .
+make
+# as root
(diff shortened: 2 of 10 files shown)


eidheim:
Regarding 1: We have gotten positive feedback on the "no-bullshit, developer oriented interface" we have, I think we should try keep it that way. That is, I'd like to keep things as they are currently, as the configuration options are pretty easy to find, and the github page is functioning as the "greeting". This way we keep things to the point (no-bullshit). Also, I think what most people will be looking for the first time is actually not the configuration options, but the new C++ project menu element. 

Regarding 2: I tried using an entry instead of a dialog to create new files, but I got confused myself what would happen when I clicked new file:) Also here the current solution is better I think. 

Regarding 3: I do not think this is needed (have thought about it previously), at least not at this point, as the developers contributing to this project most likely have of C++ programming and open source experience already, and know these things. At least that is my experience from other C++ projects. If it was a Java or Javascript project, I would agree completely! Finally, such documents may feel somewhat strict when reading them, and might scare people away from contributing to the project. It is better to comment in the PRs on this, in the very few cases this will be a problem.

eidheim:
Thumbs up for v1.0.0 release:)

eidheim:
See #118. Merge this before release when all the checkpoints are done.

PR #118, opened 2015-12-14
Correct next/previous tab keys for Linux

eidheim:
If this works better for your Arch Linux we can merge this before v1.0.0 release.

Tested on:
- [x] Debian testing
- [x] Debian stable
- [ ] Linux Mint
- [ ] Ubuntu 14
- [x] Ubuntu 15
- [x] Arch Linux

### DIFF
--- a/src/files.h
+++ b/src/files.h
@@ -98,8 +98,8 @@ const std::string configjson =
 "        \"kill_last_running\": \"<primary>Escape\",\n"
 "        \"force_kill_last_running\": \"<primary><shift>Escape\",\n"
 #ifdef __linux
-"        \"next_tab\": \"<primary>Page_Down\",\n"
-"        \"previous_tab\": \"<primary>Page_Up\",\n"
+"        \"next_tab\": \"<primary>Tab\",\n"
+"        \"previous_tab\": \"<primary><shift>Tab\",\n"
 #else
 "        \"next_tab\": \"<primary><alt>Right\",\n"
 "        \"previous_tab\": \"<primary><alt>Left\",\n"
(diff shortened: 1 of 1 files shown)


zalox:
I don't like these hot-keys. From other editors I'm used to ctrl+tab to change tabs and ctrl+shift+tab to go backwards. This is also how it's done in firefox, chrome etc.

eidheim:
It is the default for all linux applications I have seen so far

eidheim:
Aha, one sec, I see you can also use ctrl-tab and shift-control-tab

eidheim:
ok I|ll change it

eidheim:
There, its fixed:)

zalox:
http://0o2471.net/63636

We need an option to disable the grid as well :P

eidheim:
Lets not try to fix too much, and thus introduce new bugs now that we have a working version. The grid is part of style options, and juci looks awful anyway with the standard ugly sourceview styles.

eidheim:
And btw, the dark style you have is the issue, not juci. The standard sourveview styles have grid settings. This is missing on the dark style. Puh, one issue gone:)

zalox:
I haven't got any VM for ubunu 14 and debian stable right now, do you?

eidheim:
Yeah, in about 30 minutes when Im back home.

eidheim:
Its not actually working on Ubuntu 14, but its fixed in Ubuntu 15. I will not make a fix for Ubuntu 14 before after v1.0.0 release, if ever, as it requires some work to do this for Ubuntu 14 and not Ubuntu 15. At least now, the key binding is working for some linux distributions, as opposed as before this commit when it did not work for any. Huge improvement! Merge and release v1.0.0?

eidheim:
@zalox ready for merge even though two items are missing, see comment above.

PR #125, opened 2015-12-15
Removed  and  as per the TODO's

voldyman:
some files has `#include <iostream>` and `using namespace std;` which were marked to be removed, this branch does the same.


eidheim:
Thank you, merging this after a quick check if we have forgotten "std::" other places in these files.

PR #134, opened 2015-12-18
Update files.h

rpavlik:
Can use `cmake --build .` to invoke whatever build tool user chose, whether make, ninja, etc.


eidheim:
Great, thank you!

PR #139, opened 2015-12-20
Install desktop file in a portable, build-dir-independent way

archshift:
Install location matches http://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

Previous process of installation would attempt to create a directory before the install process, preventing usage in something like the Arch User Repository.
This also removes the need to workaround any bugs present with desktop file installation tools.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,8 +9,7 @@ project (${project_name})
 
 add_subdirectory("src")
 
-find_program(XDG_DESKTOP_MENU_EXECUTABLE xdg-desktop-menu)
-if(XDG_DESKTOP_MENU_EXECUTABLE)
-  file(MAKE_DIRECTORY "/usr/share/desktop-directories") #Workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=730621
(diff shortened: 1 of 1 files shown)


eidheim:
Thank you, tested it on Debian stable/testing and it works. On Ubuntu 14 one has to relog to update some application cache, but this is an Ubuntu bug. Relog will also fix the "Unknown Application.." menu bug, as currently Ubuntu's gtk menu proxy needs some work, especially in regards of supporting the new gtk menu system (the old is depricated). 

By the way, do you know of a way to install icons (scalable .svg for instance) correctly? I tried it some weeks back, but never figured out how it worked, especially the paths and how to update the icon caches. No rush though, we do not have an icon yet.

archshift:
According to [the icon theme spec](http://standards.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html), such a SVG file would be installed to /usr/share/pixmaps.

eidheim:
And I guess then that /usr/local/share/pixmaps works as well, and one could use `DESTINATION ${CMAKE_INSTALL_PREFIX}/share/pixmaps` when installing.

archshift:
Yes, that should work fine.

eidheim:
Thank you again, will try it when we get an acceptable icon:)

PR #144, opened 2015-12-21
Add preferences to menu

simenbw:
Added preferences to the File menu

### DIFF
--- a/src/menu.cc
+++ b/src/menu.cc
@@ -100,6 +100,11 @@ Menu::Menu() {
   "          <attribute name='action'>app.save_as</attribute>"
              +accels["save_as"]+ //For Ubuntu...
   "        </item>"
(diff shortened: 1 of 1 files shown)


eidheim:
Thank you, but its already here: https://github.com/cppit/jucipp/blob/master/src/menu.cc#L44
You'll find it somewhere I think, either on top menu (called app menu if I remember correctly), or under juCi++ in the "normal menu"/window menu.

zalox:
Great! @eidheim, we should add this before 1.0.1.

zalox:
I should stop reviewing on my phone.

eidheim:
Did you find the preference menu item? If not, what platform do you use?

simenbw:
Found it, was on the top "App menu"

PR #148, opened 2015-12-23
Better support for Allman bracket style (break before bracket)

eidheim:
Now indent left after '{', so that you do not need to use shift-tab when writing:
```c++
if(true)
{
  //do something
}
```

### DIFF
--- a/src/source_clang.cc
+++ b/src/source_clang.cc
@@ -605,6 +605,35 @@ bool Source::ClangViewParse::on_key_press_event(GdkEventKey* key) {
     get_source_buffer()->end_user_action();
     return true;
   }
+  //Indent left when writing { on a new line after for instance if(...)\n...
+  else if(key->keyval==GDK_KEY_braceleft) {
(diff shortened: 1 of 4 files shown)


eidheim:
If there are any other issues when writing with Allman bracket style, please let me know.

eidheim:
Ready for merge

src/source_clang.cc
```diff
@@ -99,8 +99,8 @@ void Source::ClangViewParse::configure() {
   }
   
   bracket_regex=boost::regex("^([ \\t]*).*\\{ *$");
-  no_bracket_statement_regex=boost::regex("^([ \\t]*)(if|for|else if|catch|while) *\\(.*[^;}] *$");
-  no_bracket_no_para_statement_regex=boost::regex("^([ \\t]*)(else|try|do) *$");
+  no_bracket_statement_regex=boost::regex("^([ \\t]*)(if|for|else if|while) *\\(.*[^;}] *$");
```
zalox: I think 'switch' also fits here.

PR #153, opened 2015-12-31
Debug integration

eidheim:
* The solution should be stable and flexible
* For debugging functions that you do not yet find in the menu, use Run Command in the Debug menu. 
* You can add and remove breakpoints also when debugging is running or is stopped
* Use Go to stop to figure out where the debugger stopped
* The kill debug menu item is to kill the debug process
* We will not force buffers to be non-editable when you run debug, so take that into consideration when strange things happen


eidheim:
Testing need for a week or so before we can consider merging into master.

mingwandroid:
I switched the MSYS2 MINGW-packages/jucipp-git/PKGBUILD to:
`source=("${_realname}"::"git+https://github.com/cppit/jucipp.git#branch=debug_integration"`
.. to quickly test this and it worked fine.

eidheim:
Thank you. Although note that currently juCi++ compiles without debugging support on MSYS2, or otherwise when the lldb library is not found. 

Your lldb terminal suggestion was splendid by the way! I'll have a look at your other suggestion as well, clickable call stacks, in the following days.

mingwandroid:
@eidheim By the way, would you mind accepting maintainership of the package? You don't need to build releases, it mostly a nominal thing in the PKGBUILD file (I can make the edit. or you can, demote me to a "Contributor", or remove me altogether, I don't mind!)

In general, we like the actual project members (or leaders!) to do this, not just as it endorses MSYS2, but also because, well, you know far more about the package than I (nearly nothing) so if people have questions about it, and ask me, I really won't be able to answer without asking you anyway.

eidheim:
@mingwandroid Sure, I have to test on MSYS2 either way before merging a PR to master. Will not do any changes to the current roles/permissions.

eidheim:
Tomorrow I'm going to implement the following:
* Clickable call stack (if I understand it this is the same as a backtrace, that is one can select to go to the various frames in a stopped debug process thread (http://lldb.llvm.org/cpp_reference/html/classlldb_1_1SBThread.html#ae4fc59e1a3c9a23ffacaa0302c3e601f). Will probably use the same selection dialog as is currently used in for instance Go To Usage, with function names maybe in bold. 
* Add "Set Run Arguments" in Project and Debug menus, where one can set the executable to be run, and its parameters. These arguments will follow the project where Compile and Run or Start Debug is activated. 
* Fix a crash  when quitting juCi++ when doing start debug and before the debug is started (while the debug build is compiling).

After this I guess the debug implementation is pretty complete.

eidheim:
Guess its nearing completion. I do not now know of any improvements I should do as of now actually, but that might change of course. 

@mingwandroid was this what you meant by clickable call stack (backtrace):

screen shot 2016-01-02 at 13 23 22

eidheim:
There is a crash on Ubuntu 14 when showing backtrace. I'll have a look at it later. Works fine on Debian testing.

eidheim:
Ok, found the issue on Ubuntu 14 (same on Debian stable). Should be fixed now.

eidheim:
A couple of issues at the moment, I'll fix it in a few hours.

eidheim:
Everything seems ok now, ready for merge I think, but lets test it a couple of days first.

@zalox which package do you need to install on Arch Linux to get the lldb library and headers?

eidheim:
Have to look at debugging in different threads, seems like the frame functions only works on the main thread at the moment. Fixing this Tomorrow.

eidheim:
Never mind the comment above, I fixed it now. Debug now works in multithreaded applications.

eidheim:
@archshift you might want to update your Arch Linux package to include the lldb package. This pull request will be merged in a few days.

zalox:
Hi I'm back on the internet! I'll test on arch now!

zalox:
I installed "lldb", built, and got this error when src/juci:
```bash
[zalox@towel build]$ src/juci 
: CommandLine Error: Option 'aarch64-reserve-x18' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```
(I also did pacman -Suy)

eidheim:
See https://llvm.org/bugs/show_bug.cgi?id=22543

eidheim:
https://www.archlinux.org/packages/extra/i686/lldb/ is flagged out of date. Hopefully they will fix this soon.

eidheim:
3.7.1 soon coming to Arch Linux: https://projects.archlinux.org/svntogit/packages.git/log/trunk?h=packages/llvm

eidheim:
@zalox have you tried this branch? We can start consider a merge after lldb is updated on Arch Linux.

eidheim:
This branch should be very stable now, and the debug integration pretty complete. However, I need some feedback from people that actually have used a debugger for other than backtaces;)

eidheim:
It is possible to get return value when step out is used in lldb. I'll add this to tooltips Tomorrow most likely.

zalox:
Hi, I'm testing the features now! I'm looking forward to start developing with this, great work!

eidheim:
Did you get it to work on Arch Linux?

zalox:
No, I'm running jessie. Hopefully I can use some AUR-packages to get it to work in arch.

eidheim:
Hopefully it's a matter of days before they fix https://projects.archlinux.org/svntogit/packages.git/log/trunk?h=packages/llvm

eidheim:
Also, keep in mind that jessie is using lldb 3.5, so there might be some issue with vectors having wrong size for instance when getting debug values/return values.

zalox:
Did you implement a way to see variable values?

eidheim:
One possibility could be to build an older llvm package from an older commit, for instance with llvm 3.6.2 on Arch Linux.

eidheim:
Yes, its in the tooltips. Also use Debug Run Command `frame variable` to see what is available in the frame.

eidheim:
And use the backtrace dialog to swap between frames if needed.

zalox:
I got some inconsistent behaviour:
```c++
// menu.cc:371:21
ui_xml.append(""); // only shows std::string on hover
 // on the line above, hovering the accels-variable works fine.
```

eidheim:
See if the ui_xml object is available in the frame: `frame variable`. Also you might want to turn off -O3 when debugging as many variable then gets optimized away.

eidheim:
Not quite sure how to get object member variables though, other than going to a frame where the object is in scope.

zalox:
You can hack the variable by doing something like this:
```c++
  std::string xml = ui_xml;
  xml;
```
And look at the xml instead.

eidheim:
Might also consider implementing Debug menu items: "show frame variables" and "show all variables" later if this is wanted.

eidheim:
Although in this case, ui_xml would never be shown, even with show all variable, you have to do the workaround with ...xml = ...

eidheim:
`auto xml = ui_xml;` should be enough though, as long as you stop after this line.

eidheim:
@zalox Would you mind try out the new show variables feature?

eidheim:
Keep in mind that lldb3.5 shows too many variables, the result is better in debian testing for instance (lldb3.6).

zalox:
Great feature! We should find a different solution for large buffers though.

eidheim:
lldb does this automatically (might depend on version, 3.6 at least is ok), but if the tooltips are too large still, one has the possibility to use the Debug run command to get the values in the terminal instead.

eidheim:
@zalox Have made some quite large cleanups, have tested it on multiple platforms but probably a good idea if you test it as well.

eidheim:
@zalox Would you mind test if this branch works with the new lldb package from Arch Linux Testing: https://www.archlinux.org/packages/?q=lldb? You probably have to upgrade all your llvm related packages (clang for instance) to this version (3.7.1).

eidheim:
@zalox llvm (and lldb) version 3.7.1 is now part of the standard Arch Linux packages: https://www.archlinux.org/packages/?sort=&q=lldb&maintainer=&flagged=

eidheim:
Error on Arch Linux still not fixed, I opened the following issue: https://bugs.archlinux.org/task/47811?project=1&cat%5B0%5D=2&string=lldb

zalox:
I'll find time to test tonight. I'm kind of swamped at the moment.

I think we can release this feature tomorrow if I don't find any hiccups.

eidheim:
Great, you can merge, but wait a bit with the release. I can write the changelog there btw and do the release. Was thinking about updating the screenshots, and maybe add two more too under the two existing (same size if that is ok). 

Also, the Arch Linux developers are working on fixing the lldb issue: https://llvm.org/bugs/show_bug.cgi?id=24953

PR #160, opened 2016-02-06
Removed redundant keybinding

hwidth:
Keybinding for compile_and_run was defined twice


eidheim:
Thank you for noticing!

PR #162, opened 2016-02-11
Project cleanup - easier to support compile/run/debug for other languages

eidheim:
* Now supports running html, python, js and md files
* Window class now contains less code, moved some code to a new class Project and its subclasses
* ~~A bit of work left, want to move the remaining debug source code from Window to a Project singleton class for instance (not yet decided best way to do this)~~

### DIFF
--- a/src/debug_clang.h
+++ b/src/debug_clang.h
@@ -1,5 +1,5 @@
-#ifndef JUCI_DEBUG_H_
-#define JUCI_DEBUG_H_
+#ifndef JUCI_DEBUG_CLANG_H_
+#define JUCI_DEBUG_CLANG_H_
 
 #include <boost/filesystem.hpp>
 #include <unordered_map>
@@ -9,7 +9,7 @@
 #include <thread>
 #include <mutex>
 
-class Debug {
+class DebugClang {
 public:
   class Frame {
   public:
@@ -31,10 +31,10 @@ public:
     int line_index;
   };
 private:
-  Debug();
+  DebugClang();
 public:
-  static Debug &get() {
-    static Debug singleton;
+  static DebugClang &get() {
+    static DebugClang singleton;
     return singleton;
   }
   
--- a/src/debug_clang.h
+++ b/src/debug_clang.h
@@ -9,75 +9,77 @@
 #include <thread>
 #include <mutex>
 
-class DebugClang {
-public:
-  class Frame {
+namespace Debug {
+  class Clang {
   public:
-    uint32_t index;
-    std::string module_filename;
-    std::string file_path;
-    std::string function_name;
-    int line_nr;
-    int line_index;
-  };
-  class Variable {
+    class Frame {
+    public:
+      uint32_t index;
+      std::string module_filename;
+      std::string file_path;
+      std::string function_name;
+      int line_nr;
+      int line_index;
+    };
+    class Variable {
+    public:
+      uint32_t thread_index_id;
+      uint32_t frame_index;
+      std::string name;
+      std::string value;
+      boost::filesystem::path file_path;
+      int line_nr;
+      int line_index;
+    };
+  private:
+    Clang();
   public:
-    uint32_t thread_index_id;
-    uint32_t frame_index;
-    std::string name;
-    std::string value;
-    boost::filesystem::path file_path;
-    int line_nr;
-    int line_index;
-  };
-private:
-  DebugClang();
-public:
-  static DebugClang &get() {
-    static DebugClang singleton;
-    return singleton;
-  }
-  
-  void start(const std::string &command, const boost::filesystem::path &path="",
-             const std::vector<std::pair<boost::filesystem::path, int> > &breakpoints={},
-             std::function<void(int exit_status)> callback=nullptr,
-             std::function<void(const std::string &status)> status_callback=nullptr,
-             std::function<void(const boost::filesystem::path &file_path, int line_nr, int line_index)> stop_callback=nullptr);
-  void continue_debug(); //can't use continue as function name
-  void stop();
-  void kill();
-  void step_over();
-  void step_into();
-  void step_out();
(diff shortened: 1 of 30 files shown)


zalox:
I will review this on Sunday. I'm currently on vacation.

eidheim:
No hurry, not quite finished with the cleanup, but you can maybe see if I've done something stupid.

eidheim:
Ok, I think I'm done with the cleanups. Guess we should wait some days before merge, though, so we can test it properly.

src/debug_clang.h
```diff
@@ -0,0 +1,85 @@
...
+#include <boost/filesystem.hpp>
+#include <unordered_map>
+#include <lldb/API/SBDebugger.h>
+#include <lldb/API/SBListener.h>
+#include <lldb/API/SBProcess.h>
+#include <thread>
+#include <mutex>
+
+namespace Debug {
+  class Clang {
```
zalox: I'd like to see more abstraction here. The methods defined in this class are useful in every debugger implementation (and which are called by the UI callbacks). If we were to expand to other debug implementations it's better to abstract now I guess.

eidheim: The abstraction is in project.h

src/terminal.cc
```diff
@@ -228,55 +204,41 @@ size_t Terminal::print(const std::string &message, bool bold){
...
+  });
 }
 
 bool Terminal::on_key_press_event(GdkEventKey *event) {
   processes_mutex.lock();
   bool debug_is_running=false;
 #ifdef JUCI_ENABLE_DEBUG
-  debug_is_running=Debug::get().is_running();
+  debug_is_running=Project::current_language?Project::current_language->debug_is_running():false;
 #endif
```
zalox: You said not to use these yourself. I prefer them though.

src/window.cc
```diff
@@ -24,11 +21,13 @@ namespace sigc {
 #endif
 }
 
-Window::Window() : compiling(false), debugging(false) {
+Window::Window() : notebook(Notebook::get()) {
   JDEBUG("start");
```
zalox: Why do we need this member?

src/debug_clang.h
zalox: I saw some more abstraction when I got further down. But I'm still a bit confused. If I were to add a debugger for a different language, or syntax highlighting for that matter, how would I approach?

eidheim: The source implementation is still in source*, so this is where syntax highlighting and completion for instance go (subclass Source::View like before). The project class is needed since you might have several file types related to a language, and one needs to figure out the project root path (even without a source file opened), so in order to implement a new project for a programming language with run and debug functions you subclass Project::Language. See https://github.com/eidheim/jucipp/blob/master/src/project.cc#L528-L542 for instance for python and javascript implementations for run. These are quite simple yet though. Javascript project implementation for instance need to search for the package.json file.

src/window.cc
eidheim: It's stored for convenience, notebook is used quite a few places in window.cc and writing Notebook::get(). instead of notebook. looked ugly. But can change it to Notebook::get() if you rather like that.

eidheim:
This PR is ready for merge. Tested on most platforms (including raspberian for fun:)

eidheim:
@zalox forgot to tag you.

eidheim:
@zalox I experienced a freeze when doing autocomplete in this branch. Investigating now.

eidheim:
@zalox the freeze is fixed, and this PR is again ready for merge.

src/menu.cc
```diff
@@ -7,7 +7,7 @@
 Menu::Menu() {
   auto accels=Config::get().menu.keys;
   for(auto &accel: accels) {
-#ifdef JUCI_UBUNTU_BUGGED_MENU
+#ifdef JUCI_UBUNTU
     size_t pos=0;
```
zalox: Can you explain what this does? I tried to define JUCI_UBUNTU to see if the xml changed, but it didn't it seems, or is it a char-set-fix?

src/project.cc
```diff
@@ -0,0 +1,556 @@
...
+    menu.actions["debug_stop"]->set_enabled();
+    menu.actions["debug_kill"]->set_enabled();
+    menu.actions["debug_step_over"]->set_enabled();
+    menu.actions["debug_step_into"]->set_enabled();
+    menu.actions["debug_step_out"]->set_enabled();
+    menu.actions["debug_backtrace"]->set_enabled();
+    menu.actions["debug_show_variables"]->set_enabled();
+    menu.actions["debug_run_command"]->set_enabled();
+    menu.actions["debug_goto_stop"]->set_enabled();
+  }
```
zalox: I don't see the need of duplicating this.
menu.actions["debug_goto_stop"]->set_enabled(!debug_status.empty());

eidheim: Agreed!

src/menu.cc
eidheim: Ubuntu's gtk menu-proxy is bugged (at least for Ubuntu 14 and 15, not tested in Ubuntu 16), so this is needed for the key bindings to work. If Ubuntu was not bugged, its key bindings would be set in the Menu::set_keys function. 

They might have fixed it however and backported the fix to 14 and 15. Have not tested this in a while.

src/project.cc
eidheim: Fixed

src/menu.cc
zalox: In the [https://developer.gnome.org/gtkmm-tutorial/stable/sec-menus-examples.html.en](gtkmm-example) the property is provided even on non-ubuntu-installs.

eidheim: The example is not taking into account that the keybindings can be changed. Setting the keybindings in the XML or using source code should not matter. On Ubuntu, just checked, changing keybindings do not take effect until restart. However, changing keybindings on non-Ubuntu systems works perfectly while running juCi++.

Issue #97, opened 2015-12-02
Support other build tools than cmake


eidheim:
Create a layer on top of the current cmake class so that we can use projects like https://github.com/rizsotto/Bear to create compile_commands.json in the future.

eidheim:
Also consider https://github.com/ARMmbed/yotta

photex:
Perhaps http://bazel.io

eidheim:
Interesting, I'll have a look

guysherman:
waf https://github.com/waf-project/waf

jaked122:
What sort of information is required to support additional project types? What information do I need to gather, is it just compile commands as they would be invoked through normal compilation? 

Or is there additional information that is helpful or desireable?

guysherman:
I think, in its simplest form you need two things:

1. A way to specify the build / build and run / clean / debug commands
2. A way to specify include dirs / cflags / etc for your linting that runs through libclang

You could use something like "external-build.json" which stores those details, then in your get_language function, I think, you could look for that file or CmakeLists.txt, and choose a Project::CMake or a Project::External accordingly.

(You can tell I was looking through the code the other day thinking how I might do this)

guysherman:
Yeah, I think that a quick-win would be to create a fairly open-ended option with essentially an <insert shell command here> approach for those that don't want to use CMake. But if you're looking at build systems, I recommend WAF, although processing the wscript from the ide could be tricky.

jaked122:
It looks like it supports generating other sorts of buildfiles. 

I imagine that make might even end up on the list eventually. It certainly appears to be able to generate Visual studio project files, Eclipse project files, and XCode project files. 

Given that Visual Studio files aren't likely to have information useful for clang, and Eclipse and XCode are super into xml, it should be fairly simple. 

I haven't delved into that part of waf's codebase yet, so it may just add hooks to call waf to handle the building itself, if that is the case, then it's likely to be like pulling teeth.

I'm looking at the waf code right now, and it looks like it might be easiest to add a build step capture hook into the waf/Build.py build.execute_build method, as it appears to resolve the commands inside of it.

Having said that, it is a horrible, likely broken way to acheive the goal, also it'd require keeping an internal copy of waf's codebase to execute in place of it, and someone would have to make sure it doens't break every time that waf updates.

[...]

PR #168, opened 2016-03-17
Fixes #6 and #97

### LINKED ISSUE #6: Right click menu in folder three
Create a right click menu in the folder three.

The menu should at least contain:
* Delete

edit: updated

eidheim:
~~Not thoroughly tested yet.~~ Edit: Seems ready for merge. Tested on most platforms.


src/directories.cc
```diff
@@ -209,6 +300,19 @@ void Directories::select(const boost::filesystem::path &path) {
  JDEBUG("end");
 }
 
+bool Directories::on_button_press_event(GdkEventButton* event) {
+  if(event->type==GDK_BUTTON_PRESS && event->button==3) {
+    Gtk::TreeModel::Path path;
```
zalox: Isn't there an enum for the mouse buttons?

src/directories.h
```diff
@@ -13,18 +13,30 @@
 
 class Directories : public Gtk::TreeView {
 public:
-  class ColumnRecord : public Gtk::TreeModel::ColumnRecord {
+  class TreeStore : public Gtk::TreeStore {
+  protected:
+    TreeStore() {}
+    
+    bool row_drop_possible_vfunc(const Gtk::TreeModel::Path& path, const Gtk::SelectionData& selection_data) const override;
+    bool drag_data_received_vfunc(const TreeModel::Path &path, const Gtk::SelectionData &selection_data) override;
```
zalox: Very minor, inconsistent placement of &

src/directories.cc
```diff
@@ -21,10 +23,87 @@ namespace sigc {
...
+    }
+    
+    boost::system::error_code ec;
+    boost::filesystem::rename(source_path, target_path, ec);
+    if(ec) {
+      Terminal::get().print("Error: Could not move file: "+ec.message()+'\n', true);
+      return false;
+    }
+    
+    for(int c=0;c<Notebook::get().size();c++) {
```
zalox: We should make notebook iterable. Cleans alot.

eidheim: Think it is best to do without. No need to support iterators when one never will use range algorithms (like std::equal) on the notebook tabs (compared to other tabs).

src/directories.h
eidheim: Fixed

src/directories.cc
eidheim: Fixed

eidheim:
This PR seems pretty stable, but we should probably test it for at least a week before merging.

src/dialogs.cc
```diff
@@ -40,14 +40,14 @@ std::string Dialog::gtk_dialog(const boost::filesystem::path &path, const std::s
   dialog.set_position(Gtk::WindowPosition::WIN_POS_CENTER_ON_PARENT);
   
   if(title=="Save File As")
-    gtk_file_chooser_set_filename((GtkFileChooser*)dialog.gobj(), path.c_str());
+    gtk_file_chooser_set_filename((GtkFileChooser*)dialog.gobj(), path.string().c_str());
   else if(!path.empty())
```
zalox: You should use generic_string() I think.

eidheim: Luckily it does not matter since UTF-8 is backwards compatible with ascii. Also, https://developer.gnome.org/gtk3/stable/GtkFileChooser.html#gtk-file-chooser-set-filename has `const char *filename`-parameter.

eidheim: Think the problem was that path_c.str() returned `const wchar*` instead of `const char*` on Windows.

eidheim:
There are some other minor cleanup needed, I'll fix that Tomorrow.

src/dialogs.cc
zalox: I see, I just noticed that the backslashes gets flipped and things look more beautiful in prints. My observation might be wrong. 

While writing this I also notice a c-style cast in the same place. I think we should use `std::<some cast type>` instead.

eidheim: Fixed

src/project.cc
```diff
@@ -35,6 +35,28 @@ void Project::save_files(const boost::filesystem::path &path) {
   }
 }
 
+void Project::on_save() {
+  if(Notebook::get().get_current_page()==-1)
+    return;
+  auto view=Notebook::get().get_current_view();
+  if(view->file_path.filename()=="CMakeLists.txt") {
+    auto build=get_build(view->file_path);
```
zalox: What about CMake-modules? For instance `.cmake` files. If I edit `FindLibClang` for instance then we might need a full re-parse.

eidheim: I think that should be doable as well, by looking for the first CMakeLists.txt in the parent directories and take it from there. I'll add it to my TODO list:)

eidheim: Fixed

src/source_clang.cc
```diff
@@ -639,7 +639,7 @@ void Source::ClangViewAutocomplete::autocomplete() {
   auto buffer=std::make_shared<Glib::ustring>(get_buffer()->get_text());
   auto iter=get_buffer()->get_insert()->get_iter();
   auto line_nr=iter.get_line()+1;
-  auto column_nr=iter.get_line_offset()+1;
+  auto column_nr=iter.get_line_index()+1;
```
eidheim: This commit actually fixes autocomplete after UTF-8 characters. For instance:
std::string str="æøå";
if(str=="æøå" && str.<- autocomplete now works here

zalox: Smooth!

eidheim:
@zalox This PR is ready for merge I think. Tested on most platforms.

zalox:
I'm reviewing now.

PR #171, opened 2016-03-25
Reopens last folder and files when starting juci without parameters. Also improved debug tooltips.

eidheim:


eidheim:
@zalox A small PR, should be safe to merge after you look through it. Tested on most platforms. Also checked that the exception handling worked on MSYS2 in this PR. I don't get why it does not work in all cases.

eidheim:
@zalox Found a way to improve the debug tooltips! Maybe do a little more testing before merging.

eidheim:
That is, class member values, for instance, are now getting value tooltips.

eidheim:
@zalox This PR seems ready for merge anyway. Tested on most platforms.

zalox:
Sorry for being late on this. I will review and merge tonight.

zalox:
Please merge when ready!

PR #99, opened 2015-12-05
New python plugin implementation

zalox:
# Features
* Write plugins in python using an API-module
* Configure integration with juCi++ usin JSON configuration
* Easy expansion of the API using lamba expressions in `python_api.cc`

# libjuci - Python API for juCi++
### libjuci.
```c++
  add_menu_element(str json) // -> (void) Builds a menu configured by 'json'
```
```json
[
    {
        "label": "_Edit",
        "menu_elements": [
            {
                "label":"_Snippet",
                "menu_elements": [
                    {
                        "label":"Insert snippet",
                        "keybinding":"<primary>m",
                        "action":"snippet.insert_snippet"
                    }
                ]
            }
        ]
    }
]

```
```c++
  get_juci_home() // -> (str) returns the path of the juci home folder
```
### libjuci.terminal
```c++
  println(str message) // -> (int) prints 'message' to terminal
```
### libjuci.editor
```c++
    erase(int, int); // -> (void) removes the text between the given offsets
    erase_line_range(int, int, int); // -> (void) Removes text on line 'line_number' between 'begin_line_offset' and 'end_line_offset'
    get_file(); // -> (str) Returns the current open file. If no file is open it returns empty a string
    get_highlighted_word(); // -> (str) returns json, empty string on error{'word': 'word','type': '-1'}
    get_line(); // -> (str) returns the line of text the caret is on
    get_line_number(); // -> (int) Returns the line which is being edited, -1 if no file is open or no line is selected
    get_line_offset(); // -> (int) Returns the number of characters the caret currently is positioned from the previous newline, 0 is the first character -1 if no file is open or no line is selected
    get_offset(); // -> (int) Returns the number of characters the caret currently is positioned from start of the file. 0 is the first character. Returns -1 if no lines or files are selected/open
    get_tab_char_and_size(); // -> (str) Returns a key-value json-object with 'tab_char' and 'size' keys
    get_text(); // -> (str) returns the text in the open editor
    get_text_range(int, int); // -> (str) Returns the text between 'begin_offset' and 'end_offset'. Returns emptystring if editor isn't focused on a file
    insert_at(int offset, str text); // -> (void) Inserts text at the given offset, caller is responsible to scroll to the insertion
    insert_at_cursor(str); // -> (void) Inserts text at the cursor
    is_saved(); // -> (bool) Returns true if the current open file is saved
    scroll_to(int line_number); // -> (void) Scrolls cursor to 'char_offset'
```

# Example plugins
## Toolsplugin 
[...]

PR #181, opened 2016-04-04
Python implementation using pybind11 and gobject introspection

zalox:


zalox:
Replaces #99

src/python_interpreter.cc
```diff
@@ -0,0 +1,81 @@
...
+
+PythonInterpreter::~PythonInterpreter(){
+  auto err=PythonError();
+  if(Py_IsInitialized())
+    Py_Finalize();
+  if(err)
+    std::cerr << std::string(err) << std::endl;
+}
+
+PythonError::PythonError(){
```
eidheim: @zalox Have not studied the changes in detail, just saw that you repeated Python in for instance PythonInterpreter and PythonError. I think its better to use namespaces such that it becomes: Python::Interpreter and Python::Error.

zalox:
Working plugins here for now: https://github.com/zalox/jucipp-plugins

zalox:
@milleniumbug Did you manage to test the plugins?

milleniumbug:
Not yet, I'm trying to. The missing Py_DecodeLocale was the immediate issue that prevented me from compiling this branch.

docs/install.md
```diff
@@ -143,7 +143,7 @@ Note that juCi++ must be run in a MinGW Shell (for instance MinGW-w64 Win64 Shel
 
 Install dependencies (replace `x86_64` with `i686` for 32-bit MSYS2 installs):
 ```sh
-pacman -S git mingw-w64-x86_64-cmake make mingw-w64-x86_64-toolchain mingw-w64-x86_64-clang mingw-w64-x86_64-gtkmm3 mingw-w64-x86_64-gtksourceviewmm3 mingw-w64-x86_64-boost mingw-w64-x86_64-aspell mingw-w64-x86_64-aspell-en
+pacman -S git mingw-w64-x86_64-cmake make mingw-w64-x86_64-toolchain mingw-w64-x86_64-clang mingw-w64-x86_64-gtkmm3 mingw-w64-x86_64-gtksourceviewmm3 mingw-w64-x86_64-boost mingw-w64-x86_64-aspell mingw-w64-x86_64-aspell-en mingw-w64-x86_64-gobject-introspection mingw-w64-x86_64-pygobject-devel mingw-w64-x86_64-python3-gobject
 ```
```
zalox: @milleniumbug You will probably need some variant of these packages as well. I needed to point to my site-packages on windows and debian. Arch was fine.

milleniumbug:
Ok, after installing necessary dependencies, setting `site-packages` in config.json to correct path, `snippet.py` works correctly, but `tools.py` doesn't seem to load at all.

zalox:
Are there no errors in the terminal?

milleniumbug:
None at all. I've verified that it doesn't load by putting some logging statements in `tools.py` (with `jucipp.terminal.println("test")`)

milleniumbug:
Okay, nevermind. Moving `snippet.py` outside and renaming `tools.py` to `snippet.py` loads the plugin, so probably I have forgotten to put it on some list.

zalox:
@eidheim I'm having trouble persisting my config.json file. Is there any mechanism that overwrites the file or anything in it?

I had the issues in msys2.

eidheim:
As long as you have the same version number in the config file, as in the files.h juCi++ will not touch the config file. However, if the versions are different, changes will be done the the old config file so that it corresponds to the new file specified in files.h.

zalox:
I had some very strange behaviour then. I'll try to reproduce.

eidheim:
Windows does some file lock if a file is opened by another process? That might have been the case, denying the write to you your config file. Although, I'm not very experienced with Windows.

milleniumbug:
I've debugged program, and it seems it stops searching for plugins on first hidden file.

```
/home/milleniumbug/.juci/plugins/snippet.py - full
snippet - stem
/home/milleniumbug/.juci/plugins/.gitignore - full
 - stem
/home/milleniumbug/.juci/plugins/__pycache__ - full
__pycache__ - stem
/home/milleniumbug/.juci/plugins/README.md - full
README - stem
/home/milleniumbug/.juci/plugins/LICENSE - full
LICENSE - stem
/home/milleniumbug/.juci/plugins/tools.py - full
tools - stem
/home/milleniumbug/.juci/plugins/.git - full
 - stem
```

Changing from 

```
    auto module_name=it->path().stem().string();
    if(module_name.empty())
      break;
```

to

```
    auto module_name=it->path().stem().string();
    if(module_name.empty())
      continue;
```

fixed the problem.

src/python_interpreter.cc
```diff
@@ -95,7 +95,7 @@ Python::Interpreter::Interpreter(){
   for(boost::filesystem::directory_iterator it(plugin_path);it!=end_it;it++){
     auto module_name=it->path().stem().string();
     if(module_name.empty())
-      break;
+      continue;
     auto is_directory=boost::filesystem::is_directory(it->path());
```
zalox: @milleniumbug I fixed it this morning. You probably pulled before.

zalox:
@eidheim Would it be possible to hook up to the auto-complete dialog somehow? While looking through the source I can't see any simple way as is...

Use-case:
Writing `for` would produce an auto-complete suggestion for a snippet as well as other completion data for `for`.

zalox:
It might be possible to do it in`ClangViewAutocomplete::autocomplete_get_suggestions` by asking i.e.`Python::Interpreter::get_suggestions`. The suggestions can be dynamically produced by clang context in plugins. If the interpreter and clang works in parallel I think it might work without a huge performance drop.

This approach isn't very generic and re-factoring might be needed for a reasonable solution.

eidheim:
@zalox I've been thinking about simpler ways to reuse the autocompletion dialog. I'll prioritize this in the coming weeks.

zalox:
I updated the docker images with python now. I only updated the packages, so it shouldn't break other builds. https://hub.docker.com/r/cppit/jucipp/builds/

eidheim:
Coincidently, I'll test it on master now:)

zalox:
https://github.com/cppit/jucipp-docker/commit/9cb5f6e2a5051342fc9b22439372ed9cb84c6c5f

PR #184, opened 2016-04-05
replaced direct lock()/unlock() calls to mutex with unique_lock

milleniumbug:
As per [`std::mutex::lock()`](http://en.cppreference.com/w/cpp/thread/mutex/lock) documentation:

> `lock()` is usually not called directly: `std::unique_lock` and `std::lock_guard` are used to manage exclusive locking. 

This is of course due to violated exception safety if any function throws.

### DIFF
--- a/src/debug_clang.cc
+++ b/src/debug_clang.cc
@@ -110,7 +110,7 @@ void Debug::Clang::start(const std::string &command, const boost::filesystem::pa
   debug_thread=std::thread([this, callback, status_callback, stop_callback]() {
     lldb::SBEvent event;
     while(true) {
-      event_mutex.lock();
+      std::unique_lock<std::mutex> lock(event_mutex);
       if(listener->GetNextEvent(event)) {
         if((event.GetType() & lldb::SBProcess::eBroadcastBitStateChanged)>0) {
           auto state=process->GetStateFromEvent(event);
@@ -178,7 +178,6 @@ void Debug::Clang::start(const std::string &command, const boost::filesystem::pa
               stop_callback("", 0, 0);
             process.reset();
             this->state=lldb::StateType::eStateInvalid;
-            event_mutex.unlock();
             return;
           }
           else if(state==lldb::StateType::eStateCrashed) {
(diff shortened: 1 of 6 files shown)


eidheim:
Thank you, I'll merge it after I have studied the changes in a few days or so.

src/debug_clang.cc
```diff
@@ -208,79 +206,71 @@ void Debug::Clang::start(const std::string &command, const boost::filesystem::pa
             Terminal::get().async_print(std::string(buffer, n), true);
         }
       }
-      event_mutex.unlock();
       std::this_thread::sleep_for(std::chrono::milliseconds(200));
```
ff2000: That will leave the mutex locked unnecessarily for 200 msec. Don't know how critical that is. Could be solved by a dedicated scope.

eidheim: That is true. This will lead to less responsive debugging actions if I'm not mistaken.

milleniumbug: ...or `lock.unlock();` where `event_mutex.unlock()` was.

eidheim:
I also see one could use atomic<bool> stop and atomic::compare_exchange_strong in terminal.cc instead of bool stop and mutex stop_mutex.

src/source_clang.cc
```diff
@@ -135,16 +138,16 @@ void Source::ClangViewParse::parse_initialize() {
                   parsed=true;
                   set_status("");
                 }
-                parse_mutex.unlock();
+                parse_lock.unlock();
```
eidheim: This one is not needed is it (line 141)?

eidheim: I removed it:)

milleniumbug: To be honest, if I wasn't 100% sure if the code would be equivalent, I took the safer route of just locking and unlocking the `std::unique_lock` (with `std::defer_lock`, which makes `std::unique_lock` not lock immediately in the constructor). This would make it 100% equivalent, with the added benefit of being exception safe.

PR #186, opened 2016-04-06
Cleanup of CMakeLists.txt

zalox:
@eidheim Please review and merge if ok!

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -25,9 +25,12 @@ if(MSYS)
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DMSYS_PROCESS_USE_SH")
 endif()
 
-INCLUDE(FindPkgConfig)
-
 find_package(LibClang REQUIRED)
+find_package(Boost 1.54 COMPONENTS thread log regex system filesystem REQUIRED)
+find_package(ASPELL REQUIRED)
+set(LIBCLANGMM_INCLUDE_DIR ../libclangmm/src)
+set(TINY_PROCESS_INCLUDE_DIR ../tiny-process-library)
+
 string(REPLACE libclang liblldb LIBLLDB_LIBRARIES "${LIBCLANG_LIBRARIES}")
 if(EXISTS "${LIBLLDB_LIBRARIES}")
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -41,129 +44,101 @@ else()
...
-  ${LIBLLDB_LIBRARIES}
-  #${PYTHON_LIBRARIES}
-)
+include_directories(${global_includes})
+add_executable(${project_name} ${project_files})
+target_link_libraries(${project_name} ${global_libraries})
 
-#  install(TARGETS ${project_name} ${module}
-install(TARGETS ${project_name}
+install(TARGETS ${application_name}
```
eidheim: Make install does not install the juci binary. 

I tried `message("${application_name}")` outputs "" on OS X (only tested here), while `message("${project_name}")` outputs "juci" correctly. I guess you intended to write `install(TARGETS ${project_name}` here or?

zalox: Yes, I'll fix it!

zalox: Fixed!

src/CMakeLists.txt
```diff
@@ -41,129 +44,101 @@ else()
...
-  ${GTKMM_LIBRARIES}
-  ${GTKSVMM_LIBRARIES}
-  ${Boost_LIBRARIES}
-  ${ASPELL_LIBRARIES}
-  ${LIBLLDB_LIBRARIES}
-  #${PYTHON_LIBRARIES}
-)
+include_directories(${global_includes})
+add_executable(${project_name} ${project_files})
+target_link_libraries(${project_name} ${global_libraries})
```
eidheim: Nice you cleaned up these two files! 

Although, the variables `global_includes` and `global_libraries` are only used once. Maybe better to just use` include_directories` and `target_link_libraries` on the various `${..._INCLUDE_DIRS}` and `${..._LIBRARIES}` directly then (like before)?

zalox: list(APPEND global_includes)

I was thinking we might need that in the future.

eidheim: Good point, I'll merge it Tomorrow morning or so after testing on all the OS'es.

PR #187, opened 2016-04-06
added new directory tree feature: create new file here

milleniumbug:
This pull request adds a new feature to the directory tree: ability to create new files with the option in the context menu.

before

context_menu

entrybox_before

entrybox_after

after

### DIFF
--- a/src/directories.cc
+++ b/src/directories.cc
@@ -309,7 +309,10 @@ Directories::Directories() : Gtk::TreeView(), stop_update_thread(false) {
         bool is_directory=boost::filesystem::is_directory(*source_path);
         auto target_path = (is_directory ? *source_path : source_path->parent_path())/content;
         if(!boost::filesystem::exists(target_path))
+        {
           filesystem::write(target_path, "");
+          Notebook::get().open(target_path);
+        }
         else
           Terminal::get().print("Cannot create "+target_path.string()+": file already exists.\n", true);
         update();
--- a/src/directories.cc
+++ b/src/directories.cc
@@ -299,7 +299,7 @@ Directories::Directories() : Gtk::TreeView(), stop_update_thread(false) {
   });
   menu.append(menu_item_delete);
   
-  menu_item_create.set_label("Create new file here");
+  menu_item_create.set_label("Create file");
   menu_item_create.signal_activate().connect([this] {
     if(menu_popup_row_path.empty())
       return;
--- a/src/directories.cc
+++ b/src/directories.cc
@@ -299,8 +299,8 @@ Directories::Directories() : Gtk::TreeView(), stop_update_thread(false) {
   });
   menu.append(menu_item_delete);
   
-  menu_item_create.set_label("Create file");
-  menu_item_create.signal_activate().connect([this] {
+  auto create_file_label = "Create file";
+  auto create_file_signal = [this] {
     if(menu_popup_row_path.empty())
       return;
     EntryBox::get().clear();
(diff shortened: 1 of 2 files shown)


eidheim:
Thank you again. I'll probably make the v1.1.2 release Tomorrow if no new issues arises, and look through these changes then. 

Just a few minor comments:
* new file item is normally before Rename and Delete?
* you can open the newly created file with Notebook::get().open(...) after the file is created. Then the new file gets highlighted too.

milleniumbug:
I thought about it a little before (if to make it on top, or on the bottom), but in my opinion, putting it on the bottom would be better since: a.) "Rename" is done more often so it makes more sense to have it on top b.) I didn't want to disturb muscle memory of people who already use the program :D

I'm not yet sure if I want the new file to be open automatically, but highlighting the file would be nice on its own.

eidheim:
I am not sure either on my own two comments, but I guess one typically wants to write in the new file, so opening it is ok I guess, and also the normal behaviour in other IDE's. Regarding b) though,I prefer looking forward and not backward, so changing the current layout is ok in my opinion. I'll leave it up to you though:)

milleniumbug:
Ok, now it's opening the file after creation, with the layout staying as-is.

zalox:
Maybe change text to `New file`. I guess the `Create` and `here` are redundant.

milleniumbug:
@zalox I'd prefer `Create file`, since this is unambiguously an action. Yup, `here` is definitely redundant, so I'll remove it.

zalox:
Create file is marvelous!

zalox:
I can't seem to create files in project root, if the project root has only folders. (if that makes sense)

milleniumbug:
Good catch, that's definitely a problem. Maybe not that big, but definitely embarrassing. Not sure how to solve it:

- treating the empty space below as if the action was done on the root directory? probably the simplest option, but it doesn't work with enough directories and files.
- add sentinel `(empty)` entry for the directories with only directories? ugly, but works for all cases.
- add a top level root in the directory tree? probably the most intuitive, but that opens doors to new bugs: renaming the root directory while it's in use, or hell, deleting it.

eidheim:
One solution could be when right clicking the empty space in directories, or when right clicking on directories title (the current opened directory). But then one would need two dropdown menus. One when right clicking on a file or directory, and one when right clicking empty space/title.

eidheim:
I also saw now that trying to move (empty) gave an error message in terminal. I've added it to my TODO.

zalox:
@milleniumbug I like yout first option. I don't fully understand your concern though?

Is it when the directory is filled with directories (all the way to the bottom)?

eidheim:
@zalox yes when the directory view is full, but then one could right click on the directory title instead as I tried to explain:)

zalox:
@eidheim and by the directory title you are describing the folder tree label on the top?

eidheim:
@zalox yes

zalox:
So we agree on this I guess:
* Right click in empty area
* Right click on title on top (solves issue with directories all the way to the bottom)

I got one final concern. Is the menu that pops up the same for all these use-cases? In case of yes, what happens if I "Remove" or "Rename" on both those use-cases?

eidheim:
@zalox See my comment above; for this we need two menus, but that is solvable.

milleniumbug:
Yup, exactly what @eidheim means. I've tried implementing the suggestion, however I've hit a wall - the empty space on the bottom is clickable (with the context menu containing only the option of creating the new file) the title bar is clickable, however, clicking on the title bar is treated as if it was done on the first element of the tree (first displayed file/directory).

zalox:
@eidheim If we move the construction from the directories construction of the menu to the right-click event, we should be able to produce a menu based on the current context instead of predefining it. My only concern is performance.

eidheim:
@zalox Both menus can be created in the constructor, no worries there. 
@milleniumbug That is strange, on both OS X and Debian testing I currently get no dialog when right clicking on the title (that is, get_path_at_pos returns false).

eidheim:
@milleniumbug I see now what you mean. Using `Gtk::TreeView::set_headers_clickable();` is problematic on my systems also.

eidheim:
@zalox and @milleniumbug I managed to show/hide menu items using for instance menu_item_rename.show(); and menu_item_rename.hide(); when appropriate. We can use this instead of creating two menus. My bad!

eidheim:
@milleniumbug I managed to fix the right click problem with for instance:
```diff
diff --git a/src/directories.cc b/src/directories.cc
index 2937374..50f28de 100644
--- a/src/directories.cc
+++ b/src/directories.cc
@@ -299,6 +299,17 @@ Directories::Directories() : Gtk::TreeView(), stop_update_thread(false) {
   
   menu.show_all();
   menu.accelerate(*this);
+  
+  set_headers_clickable();
+  forall([this](Gtk::Widget &widget) {
+    if(widget.get_name()=="GtkButton") {
+      widget.signal_button_press_event().connect([this](GdkEventButton *event) {
+        if(event->type==GDK_BUTTON_PRESS && event->button==GDK_BUTTON_SECONDARY && !path.empty())
+          cout << "show no_path_menu" << endl;
+        return true;
+      });
+    }
+  });
 }
 
 Directories::~Directories() {
@@ -398,6 +409,11 @@ bool Directories::on_button_press_event(GdkEventButton* event) {
     if(get_path_at_pos(static_cast<int>(event->x), static_cast<int>(event->y), path)) {
       menu_popup_row_path=get_model()->get_iter(path)->get_value(column_record.path);
       menu.popup(event->button, event->time);
+      cout << "show path_menu" << endl;
+      return true;
+    }
+    else if(!this->path.empty()) {
+      cout << "show no_path_menu" << endl;
       return true;
     }
   }
```

milleniumbug:
Cool! I'll check it out when I get home.

PR #194, opened 2016-04-15
changed main() created by "New C++ project" option

milleniumbug:
removing `using namespace std;` and usage of `endl`, replacing it with `\n`

Usage of `using namespace std;` is [questionable](https://stackoverflow.com/questions/1452721/why-is-using-namespace-std-in-c-considered-bad-practice) and `std::endl` is a badly designed manipulator - it does a different thing that the name indicates: it outputs a newline *and* flushes the output. I've taken the liberty of removing `return 0;`, although I don't mind it that much.

Why does it matter for the IDE? Even though it's a minor point, the sample created by IDE should be *spotless*, as this is the first thing the Random Joe sees when he creates new project, it influences the first experience with the new IDE, so we don't want to encourage bad style in the samples. Also many other C++ IDEs fail the "Hello World test" - we can do better.

### DIFF
--- a/src/window.cc
+++ b/src/window.cc
@@ -218,7 +218,7 @@ void Window::set_menu_actions() {
         return;
       }
       std::string cmakelists="cmake_minimum_required(VERSION 2.8)\n\nproject("+project_name+")\n\nset(CMAKE_CXX_FLAGS \"${CMAKE_CXX_FLAGS} -std=c++1y -Wall -Wextra -Wno-unused-parameter\")\n\nadd_executable("+project_name+" main.cpp)\n";
-      std::string cpp_main="#include <iostream>\n\nusing namespace std;\n\nint main() {\n  cout << \"Hello World!\" << endl;\n\n  return 0;\n}\n";
+      std::string cpp_main="#include <iostream>\n\nint main() {\n  std::cout << \"Hello World!\\n\";\n}\n";
       if(filesystem::write(cmakelists_path, cmakelists) && filesystem::write(cpp_main_path, cpp_main)) {
         Directories::get().open(project_path);
         notebook.open(cpp_main_path);


eidheim:
I see your points, but the example is mainly targeting new users. I'll have a hard time convincing my new C++ students that writing `std::` in front of every function is good practice. If I leave it to later, when working with header and implementation files, it makes more sense. 

`std::endl` is also great for beginners that are used to other languages, since for instance System.out.println also flushes the standard output. Additionally, `std::endl` corresponds to `\r\n` (if I'm not mistaken) and flush on Windows. 

However, I'm open for discussion!

milleniumbug:
I'm not convinced by your first argument. You have to explicitly introduce `using namespace std;` in order to not write `std::`. IME as a student, most teachers handwave this away with "yeah, this is what you need to use `std`", which is bad from educational standpoint. (another possibility is that they're basing on pre-standard C++ materials, which makes introducing `using namespace std;` a "quick fix" to make them work) Why is it bad? Because you have to make them unlearn the habit later - some of them do unlearn, and some others don't - and those who don't write `using namespace` in header files which is simply wrong.

(Fun fact: I used to write `using namespace std;` too, but only introduction to header files and separate compilation made me stop because at one place (declarations) you need to write fully qualified name, and on the other, you don't (definitions) - that was the simple thing that put me away, not problems with ADL, shadowing issues or anything else)

Wrt `std::endl`: actually, doing `<< std::endl` is equivalent to `<< "\n" << std::flush`, so no portability differences there. Whether writing `"\n"` counts as writing `\n` to the output stream or writing implementation-defined newline sequence (that is, `\r\n` on Windows and `\n` on Linux) is dependent on if the stream is opened in binary mode or text mode. (`std::cout` is opened by default in text mode)

27.7.3.8 C++11 N3337 draft

> Effects: Calls os.put(os.widen(’\n’)), then os.flush().

(`.widen()` is a member function that returns a widened char, in case of `std::ostream` it returns a `char`)

> since for instance `System.out.println` also flushes the standard output.

Actually I'm not sure about that, [can't find it](https://docs.oracle.com/javase/7/docs/api/java/io/PrintStream.html#println%28%29) in the Java docs (although they do make explicit that it writes `line.separator`)

eidheim:
Actually, Bjarne Stroustrup agrees with you (http://www.stroustrup.com/hello_world.c), so I have no choice but to merge this;)

PR #196, opened 2016-04-19
Fixed minor memory leak

milleniumbug:
Nuke raw calls to `new` from orbit, any one of them becomes a potential to leak.

I'm seriously considering backporting `make_unique` and replacing all calls to `new` with it.

### DIFF
--- a/src/project_build.cc
+++ b/src/project_build.cc
@@ -2,9 +2,9 @@
 #include "config.h"
 
 std::unique_ptr<Project::Build> Project::get_build(const boost::filesystem::path &path) {
-  auto cmake=new CMake(path);
+  std::unique_ptr<Project::Build> cmake(new CMake(path));
   if(!cmake->project_path.empty())
-    return std::unique_ptr<Project::Build>(cmake);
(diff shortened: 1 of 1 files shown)


nabijaczleweli:
10/10 would merge

eidheim:
Good catch. Looking forward moving from c++11 to c++14! Of the larger distributions, only Ubuntu 14 (still the recommended version) is holding us back I think.

vadimgit:
> Looking forward moving from c++11 to c++14!

Ubuntu 16.04 is coming out soon. Good time to switch to c++14, std::make_unique for example helps to catch errors like this at compile time. It's really forces to design project or class in more robust way from the very beginning.

eidheim:
@vadimgit I'm all for dropping support for Ubuntu 14 when that happens, and when Ubuntu 15 becomes the recommended Ubuntu distribution (hopefully at the same time). Maybe we can get rid of the ugliness found for instance in https://github.com/cppit/jucipp/blob/master/src/source.cc#L11-L23 then too.

PR #197, opened 2016-04-25
cleanup: removal of trailing whitespace and tabs

zalox:
All trailing whitespace is removed and tabs are replaced with spaces.


eidheim:
Not 100% sure about this PR, that is on the removal of trailing whitespaces. Is trailing whitespaces a problem? I think we should not force people to remove trailing whitespaces in their PRs, as this is a requirement that is not really needed. And cleanup of such PRs, to keep removing trailing whitespaces, is extra work and lead to commits without any real value. 

Also, I'd like to keep my cleanup_whitespace_characters setting (in preferences) to false, since I'm working on many different projects. Turning this setting to true would sometimes, depending on the project, lead to large commits such as this one.

zalox:
@eidheim I would like to discuss this again.

Since juCi++ provide functionality to clean code, it is imperative to arrange juCi++' source in the same way. If developers are using juCi++ to develop this repository, we now "disable" the `format code` feature in the development of existing files. In my opinion, we should encourage the usage of features, not the opposite.

I realise this would mean a huge commit, similar to the ones already in this PR, but It would reenable users to use clang-format instead of having to revert their changes.

eidheim:
You might be right about this. I'm probably biased since I'm set in my old ways of doing things. However, if we decide to do this, we need to add a .clang_format file with appropriate options to the repository path (we could discuss this in a pull request). Not sure if we should enforce using clang-format before accepting a commit though, and we would end up with some whitespace/clang-format commits from time to time. Again, not sure if this is a good solution. Clang-format is a bit slow too on larger files, it is not something you would want to do for every save. And one would have to think about this before commiting, and that would lead to yet another thing to keep in mind, instead of focusing on not adding bugs/crashes in the given commits.

zalox:
> we need to add a .clang_format file with appropriate options to the repository path

I agree that we need to define a `.clang_format` file, but the path we need to discuss further in the PR.

> Not sure if we should enforce using clang-format before accepting a commit though, and we would end up with some whitespace/clang-format commits from time to time.

We shouldn't enforce clang-format for contributors, it is a job for robots. I think it could be done with hooks on merge, the merge would read: "merged branch `contributor/fixbranch`  with automatic code formatting".
* original commits are untouched.
* a clang-format is performed on the merge commit
* only the merge commit would contain format cleanups, and it can be done conditionally if the format is ok.

I would not invoke cleanup on save either, it would break our simplicity principle.

eidheim:
Hooks on merge sounds great! Did not know this was an option at all. If we can do this properly I'm all for it. I'll try read up on this during the Weekend, so I too know something about this:)

zalox:
I call that a merge hook might be too invasive. We might end up with merge conflicts we don't understand. It might be more useful to clean before releases. In that way, we won't be bothered by weird merge conflicts during development. I haven't tried the release functionality on Github, but I think it's possible to do some processing in that step.

PR #203, opened 2016-04-30
make get_build generic and safety

vadimgit:
get_build was creating new Build instance with path.empty() = true  by using its default constructor . if we  use path.empty() for new CMakeBuild instance  we will get the same result and don't need to mess with type hierarchy(lCMake and Project::CMake was different classes., Project::Build and Project::CMake castings are dangerous too).   It will be great to provide the same functionality for MakeBuild, SconsBuild and so on in the future. And all internals in make.h, scons.h  the same way as cmake.h.


eidheim:
Thank you, but the point of `Project::get_build` is to deduce the build management system when other management systems than CMake gets supported. The build might not be C/C++ related either. One do not want to specify build management system type when calling Project::get_build. For instance:
https://github.com/cppit/jucipp/pull/203/files#diff-00b5229d483ad0c01de611ff630723beL175
is better than
https://github.com/cppit/jucipp/pull/203/files#diff-00b5229d483ad0c01de611ff630723beR175

By the way, what do you mean by "Project::Build and Project::CMake castings are dangerous"?

I agree that Project::CMake should be called Project::CMakeBuild. And I guess Project::Language's subclasses should be renamed similarly too.

vadimgit:
I took a look at previous pull request. And It looks like the line:
```C++
return std::unique_ptr<Project::Build>(new Project::Build());
```
 is not solve the problem but make code complicated.
If you decide to add new build tool, you will need to rewrite a lot of code in progect.cc and source.cc for example.So it does make this function "generic", but make complicated other stuff. If I would make a decision I would make something like this:

```C++
template <typename T>
class Build<T> 
```
where T is a build tool for example CMakeBuild and we will need template specialization only for creating a Build<T> instance.

```C++
auto cmake_build = std::make_unique<Build<CMakeBuild> >; 
```
the benefit is that it will force to implement the same interfaces for new build tool or error at compile time. But even this decision I am not consider like the best. There is not a lot of  "build tools" (make, ninja, cmake, autotools, scons). So I would just make Build class as abstract(interface) class and don't complicate code with templates at this level (no fancy compile errors). Build class should be the interface for build tools and nothing more. But this is all IMHO. But this is all IMHO. I can be mistaken.

eidheim:
To support build management systems other than CMake the main changes would be to the `src/project_build.*` files. The CMake reference in `Project::get_language` is only currently to identify clang projects. Furthermore, one might want add something in `Project::on_save` as well to automate the build further, like what is done when saving CMakeLists.txt. There are no references to `Project::CMake` in `src/source.cc`. 

Additionally, one would have to update `Project::get_build` in order to identify the build management system in the given directory structure, and return the correct `Project::Build` subclass object (of for instance the types `Project::CMakeBuild`, `Project::AutotoolsBuild` or `Project::SconsBuild`). It has to be subclass object, since it must be identified at runtime. 

Ninja or make files are currently created in the build directory based on the cmake parameters. This would be the same if autotools or scons for instance would be supported I guess, although I'm not very familiar with these tools.

vadimgit:
Oh yes. I understood an idea, I don't understand your approach. I just trying to say that you can avoid run-time polymorphism, just make Project::Build "pure" abstract class. I mean you don't need to iterate through build tools for example, or support many build tools in one project. Just keep code as simple as possible.

And this should be enough:
```C++
if(path.filename()=="CMakeLists.txt")
    auto build = std::make_unique <CMakeBuild>*(path);
else if((path.filename()=="SConstruct")
    auto build = std::make_unique <SConsBuild>*(path);
```

and add one method to check which build tool is using, if the logic is not generic.
```C++
auto build_type = build.type();
```

IMHO it should be much simpler, and less error prone in the future.

For example right now if get_build() returns base class instance Build() but not CMake() you do all these checks anyway.

PR #205, opened 2016-05-02
Fix me

vadimgit:
little fix


src/source_clang.cc
```diff
@@ -622,8 +622,8 @@ void Source::ClangViewAutocomplete::autocomplete_dialog_setup() {
 
 void Source::ClangViewAutocomplete::autocomplete_check() {
   auto iter=get_buffer()->get_insert()->get_iter();
-  if(iter.backward_char() && iter.backward_char() && (get_source_buffer()->iter_has_context_class(iter, "string") ||
```
eidheim: I think the reason behind the double backward_char's is because to get the correct gtksourceview's iter_has_context_class. iter.backward_char() actually moves the iter backwards as well as returns true if the iter is still valid.

vadimgit: Oo. you right. there is a reason.

eidheim:
Thanks for looking through the source though!

PR #206, opened 2016-05-03
Add doxygen configuration

goto40:
- Doxygen configuration added (see apidoc.md).
- The egenrated API doc contains a search function.
- The idea is to successive add documentation to the code and the underlying architecture:
  - static information can be added via standard doxygen comments (responsibilities of individual classes/functions/modules)
  - dynamic behaviour can be added using the plantuml syntax (e.g. activity diagrams)


eidheim:
Thank you. I'm busy till Thursday morning, but can look at this then.

docs/apidoc.md
```diff
@@ -0,0 +1,16 @@
+# juCi++ API doc
+
+## Prerequisites:
+ * doxygen
+ * plantuml (install via apt-get or download from http://de.plantuml.com/; if downloaded either copy the jar file to /usr/bin or set the environment variable PLANTUML_PATH to point to the path containing the jar file)
+
```
zalox: Notes:
* Link to install instructions provided by platuml instead. java also dependency?
* plantuml generates uml, not needed for api-only build.

src/CMakeLists.txt
```diff
@@ -141,3 +141,21 @@ target_link_libraries(${project_name} ${global_libraries})
 install(TARGETS ${project_name}
   RUNTIME DESTINATION bin
 )
+
+find_package(Plantuml)
+if(PLANTUML_FOUND)
+#message("plantuml found PLANTUML_JARFILE=${PLANTUML_JARFILE}")
+else(Plantuml_FOUND)
+message("plantuml not found. No UML diagrams are generated. You may define the environment variable PLANTUML_PATH...")
```
zalox: What does this do?

docs/apidoc.md
goto40: Yes, java is also a dependency to run plantuml. Plantuml allows to specify uml diagrams within doxygen comments.
Without plantuml, the API doc will omit the UML diagrams.

src/CMakeLists.txt
goto40: with find_package(PlantUml) plantuml is detected and later inserted into the Doxyfile (generated form Doxyfile.in). 
As noted above, if it is not detected UML diagrams will just be omitted.

docs/apidoc.md
goto40: I think with ubuntu 16.04 "apt-get install plantuml" just does the thing (I will check this).

src/CMakeLists.txt
zalox: I see, so doxygen calls plantuml?

goto40: Especially behaviour diagrams may help to provide graphical information how different components work together (e.g. using activity diagrams). See: http://goto40.github.io/create_anaglyph/page_architecture.html ("Runtime View")

goto40: Yes, doxygen integrates plantuml...

eidheim:
I think this might help new contributors a lot in understanding the source. I'm not a doxygen expert, but can we set HAVE_DOT to YES (https://github.com/goto40/jucipp/blob/add_doxygen_config/src/Doxyfile.in#L2114)? That enabled some other useful diagrams.

goto40:
I added HAVE_DOT=YES and also enabled caller graphs (showing who is calling a particular function).

Issue #158, opened 2016-02-03
[Help]I am in the link problem, My env is in ubuntu 15.04


timloo0710:
I follow the install.md. 

in the "sudo make install", 

There are errors in linking:

Linking CXX executable juci
CMakeFiles/juci.dir/juci.cc.o: In function `Application::on_command_line(Glib::RefPtr<Gio::ApplicationCommandLine> const&)':
/home/timloo/CPP_EXS/jucipp/src/juci.cc:(.text+0x51): 未定義參考到「Glib::set_prgname(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)」
...............................................................................

I add a little code in the 
src/CMakeLists.txt
........................................
link_directories(
  ${GTKMM_LIBRARY_DIRS}
  ${GTKSVMM_LIBRARY_DIRS}
  ${Boost_LIBRARY_DIRS}
  ${LIBCLANG_LIBRARY_DIRS}
)
#----- test 
message(STATUS "GTKMM_LIBRARY_DIRS: ${GTKMM_LIBRARY_DIRS} ")
message(STATUS "GTKSVMM_LIBRARY_DIRS: ${GTKSVMM_LIBRARY_DIRS} ")
message(STATUS "Boost_LIBRARY_DIRS: ${Boost_LIBRARY_DIRS} ")
message(STATUS "LIBCLANG_LIBRARY_DIRS: ${LIBCLANG_LIBRARY_DIRS} ")
................................................
and 

cmake ..

It shows

-- GTKMM_LIBRARY_DIRS:  
-- GTKSVMM_LIBRARY_DIRS:  
-- Boost_LIBRARY_DIRS: /usr/lib/x86_64-linux-gnu 
-- LIBCLANG_LIBRARY_DIRS:  
 

In my env, three libs are not found,  is there any help? Thanks!!

eidheim:
I have seen the same problem before when someone was trying to compile juCi++ with clang++ instead of g++. It is important to compile with g++ since the libraries you use are also compiled with g++ (ABI differences most likely). Would you mind post the entire output of `cmake ..`?

eidheim:
To compile juCi++ with g++ instead of clang++ (if this is the problem), replace the cmake command in the install docs with:
`cmake -DCMAKE_CXX_COMPILER=g++ ..`

timloo0710:
Thank you very much!!

I append following two in the CMakeLists.txt:

[...]

PR #207, opened 2016-05-05
Enable Continuous Integration with Travis CI

milleniumbug:
For now it doesn't do much (only builds it), but in the future it could run tests, generate packages and so on.

To enable this you need to connect the GitHub account to the free version of Travis CI on https://travis-ci.org/ - once you register there, you can enable CI for specific repositories.

### DIFF
--- a/ci/fedora-23.build
+++ b/ci/fedora-23.build
@@ -1,0 +1,6 @@
+#!/bin/bash
+cd /home &&
+mkdir build &&
+cd build &&
+cmake -DCMAKE_CXX_COMPILER=g++ .. &&
+make
--- a/ci/ubuntu-16.04.docker
+++ b/ci/ubuntu-16.04.docker
@@ -6,9 +6,9 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get update \
       cmake \
       make \
       g++ \
-      libclang-3.5-dev \
-      liblldb-3.5-dev \
-      clang-format-3.5 \
+      libclang-3.6-dev \
+      liblldb-3.6-dev \
+      clang-format-3.6 \
       pkg-config \
       libboost-filesystem-dev \
       libboost-regex-dev \
--- a/ci/ubuntu-16.04.docker
+++ b/ci/ubuntu-16.04.docker
@@ -1,4 +1,18 @@
 FROM ubuntu:16.04
 
-RUN apt-get update
-RUN apt-get install -y git cmake make g++ libclang-3.5-dev liblldb-3.5-dev clang-format-3.5 pkg-config libboost-filesystem-dev libboost-regex-dev libgtksourceviewmm-3.0-dev aspell-en libaspell-dev
+RUN DEBIAN_FRONTEND=noninteractive apt-get update \
+&& apt-get install -y \
+      git \
+      cmake \
+      make \
+      g++ \
+      libclang-3.5-dev \
+      liblldb-3.5-dev \
+      clang-format-3.5 \
+      pkg-config \
+      libboost-filesystem-dev \
+      libboost-regex-dev \
+      libgtksourceviewmm-3.0-dev \
+      aspell-en \
+      libaspell-dev \
+&& rm -rf /var/lib/apt/lists/*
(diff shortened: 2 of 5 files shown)


eidheim:
Great. Only problem is that Travis CI is running Ubuntu 14 I think, which we might stop supporting in some months in order to move to c++14 and std::regex. Do you know if they have plans to upgrade to Ubuntu 16?

eidheim:
A few thoughts on extending the Travis CI:
* compile with -Werror added to CXX extra argument environment variable
* add `scan-build` before `make` (http://clang-analyzer.llvm.org/scan-build.html) with appropriate parameters and report error if issues are reported (for instance if html pages are created)
* compile with both g++ and clang++ (although without clang++ linking, see #158)

milleniumbug:
The official answer seems to be: ["not this year, try using Docker containers instead"](https://github.com/travis-ci/travis-ci/issues/5821).

eidheim:
I guess docker containers is the way to go then, with Debian testing image or something similarly stable and new.

milleniumbug:
Sorry for the commit spam, but I didn't know a better way of testing the changes on Travis. 

This commit introduces container based setup - there are two containers set up there, with build scripts and dockerfiles put into `ci` directory.

ci/ubuntu-16.04.docker
```diff
@@ -0,0 +1,4 @@
+FROM ubuntu:16.04
+
+RUN apt-get update
+RUN apt-get install -y git cmake make g++ libclang-3.5-dev liblldb-3.5-dev clang-format-3.5 pkg-config libboost-filesystem-dev libboost-regex-dev libgtksourceviewmm-3.0-dev aspell-en libaspell-dev
```
zalox:
```Dockerfile
RUN DEBIAN_FRONTEND=noninteractive apt-get update \
&& apt-get install -y \
      git \
      cmake \
      make \
      g++ \
      libclang-3.5-dev \
      liblldb-3.5-dev \
      clang-format-3.5 \
      pkg-config \
      libboost-filesystem-dev \
      libboost-regex-dev \
      libgtksourceviewmm-3.0-dev \
      aspell-en \
      libaspell-dev \
&& rm -rf /var/lib/apt/lists/*
```

zalox: When doing two run statements with apt-get we risk cache invalidation. The one I posted makes sure that doesn't happen.

eidheim:
Thank you! This I think will form a good basis. I'll have a look at the scripts in the coming days, but we can probably merge this soon anyway. @zalox you should probably have a look at what should be done at the Travis CI side, and maybe have a look at these scripts too.

I have since this PR created automatic docker builds for ubuntu 14, debian stable, debian testing, debian unstable and archlinux that I now test all changes on before merge. Testing on various Linux platforms at least have become much easier after learning this. It's even possible to test the GUI with `docker exec debian_testing juci` for instance after a build (although this can't be done on Travis CI of course).

ci/ubuntu-16.04.docker
```diff
@@ -0,0 +1,18 @@
...
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get update \
+&& apt-get install -y \
+      git \
+      cmake \
+      make \
+      g++ \
+      libclang-3.5-dev \
+      liblldb-3.5-dev \
+      clang-format-3.5 \
```
eidheim: Can use the libclang/liblldb/clang-format 3.6 version instead. Currently this is the default llvm version in debian testing/unstable.

ci/fedora-23.build
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+cd /home &&
+mkdir build &&
+cd build &&
+cmake -DCMAKE_CXX_COMPILER=g++ .. &&
```
eidheim: Both .build files: to exit with error code if any warnings you can replace this line with `cmake -DCMAKE_CXX_FLAGS="-Werror" -DCMAKE_CXX_COMPILER=g++ .. &&` if I'm not mistaken.

eidheim:
@zalox I leave merging of this PR up to you, since you might want to study the Travis CI server side.

zalox:
I have some suggestions. I'm trying a docker-compose approach, that makes it very easy to test ci locally.

PR #208, opened 2016-05-05
Removed old Python API

eidheim:
@zalox merge if it's ok to remove these files


zalox:
To avoid conflicts I think It's better to cherry pick 1cb806eb0067482e6f7397731e90320c30823299.

zalox:
I can do it if you'd like :)

eidheim:
Sure, thank you!

PR #209, opened 2016-05-05
Remove old files

zalox:


zalox:
I squashed the previous commit into one aswell. It was kind of messy.

zalox:
@eidheim Now it should be fixed.

eidheim:
My branch is now: "This branch is 10 commits ahead, 3 commits behind cppit:master." but I can fix that.

eidheim:
Or, hmm, the ahead and behind are conflicting...

zalox:
I think that's just fixed when you do "git pull cppit master"

zalox:
Doing:
```bash
git clone https://github.com/eidheim/jucipp.git eidheim # or just use your repo
cd eidheim
git remote add cppit https://github.com/cppit/jucipp.git
git pull cppit master
git push origin master
```
should work

eidheim:
I still got "This branch is 11 commits ahead of cppit:master." but its only commit messages. What should I do:)?

zalox:
```bash
git reset 6dac0b723bb3502e11e9fd310d9e234ba02b6e05 --hard
git pull https://github.com/cppit/jucipp master
git push origin master --force
```
origin should is your repo.

zalox:
When I'm force updating I usually do a 
```bash 
$ git remote -v
origin	git@github.com:eidheim/jucipp.git (fetch)
origin	git@github.com:eidheim/jucipp.git (push)
```
To see what I'm doing.

eidheim:
Thank you, its even now. It's a bit hard to follow on the changes when these things happen (this is more important than having a couple extra commit messages in my view), and we are somewhat spamming the people who are watching the repository. On the other hand, I understand that is might be needed some times.

eidheim:
Or never mind on following the changes comment above. I see them in: https://github.com/cppit/jucipp/commit/77023ac5cbdf083c469613db4e97f2e0d0cba65f

zalox:
I agree, this got to messy, to fast. Changes like this should be done before merge, and I won't do fixes  in the future unless agreed upon. Simply asking each other to merge would stop this from happening in the first place I suppose.

eidheim:
@zalox In my opinion, we should not be very strict with the number of large or small commits in a pull request. Contributors to C++ projects are, in my experience, better than average programmers, but maybe not as experienced in git as younger contributors might be. Some are maybe not experience in git at all. If we can simplify this by less strict rules, and focus on having a good source base, I'm happy:)

That said, I try to make my own changes in as few commits as I can now. I've learned a bit more "advanced" git commands, but if I have pushed to cppit master, I try not to revert it in any way. Usually I catch smaller issues earlier though, and thus avoid commits like: "fixes to last commit" like I did earlier!

eidheim:
One additional advantage in not messing with the commits, is that the contributors to the various commits are not changed. The four last commits to cppit master Today for instance, does not include my name, even though I know I did some changes! I'm not picky about this at all, but someone might rightly be.

PR #213, opened 2016-05-06
Added: Go to Implementation

eidheim:
Works if implementation file is opened and parsed. Header files are not considered, use Go to Declaration instead then. Still pretty useful.

### DIFF
--- a/src/source_clang.cc
+++ b/src/source_clang.cc
@@ -922,19 +922,20 @@ Source::ClangViewAutocomplete(file_path, language) {
   get_implementation_location=[this](const Token &token){
     Offset location;
     if(parsed && this->language && this->language->get_id()!="chdr" && this->language->get_id()!="cpphdr") {
-      for(auto &a_token: *clang_tokens) {
-        auto cursor=a_token.get_cursor();
+      for(auto token_it=clang_tokens->rbegin();token_it!=clang_tokens->rend();++token_it) {
+        auto cursor=token_it->get_cursor();
         auto kind=cursor.get_kind();
-        if((kind==clang::CursorKind::CXXMethod || kind==clang::CursorKind::Constructor || kind==clang::CursorKind::Destructor) &&
-           a_token.get_kind()==clang::Token_Identifier && cursor.has_type()) {
+        if((kind==clang::CursorKind::FunctionDecl || kind==clang::CursorKind::CXXMethod ||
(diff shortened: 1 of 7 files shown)


eidheim:
@zalox would you mind have a quick look at this? I'll probably cleanup the refactoring methods at some point later on.

eidheim:
Also, lets say I want to combine these 3 commits into one, without opening another pull request, how would I do that?

eidheim:
@zalox I think I managed to merge the 3 commits into 1, but double check it for me please:)

src/source_clang.cc
```diff
@@ -919,6 +919,29 @@ Source::ClangViewAutocomplete(file_path, language) {
...
+           a_token.get_kind()==clang::Token_Identifier && cursor.has_type()) {
+          auto referenced=cursor.get_referenced();
+          if(referenced && static_cast<clang::CursorKind>(token.type)==referenced.get_kind() &&
+             token.spelling==a_token.get_spelling() && token.usr==referenced.get_usr()) {
+            location.file_path=cursor.get_source_location().get_path();
+            auto clang_offset=cursor.get_source_location().get_offset();
+            location.line=clang_offset.line;
+            location.index=clang_offset.index;
+            //do not break here, choose last occurrence in case declaration and implementation is in the same file
+          }
```
zalox: Why not iterate backwards?

eidheim: Good idea!

zalox: It might be a good idea to communicate to the user why a jump fails when the implementation file is closed.

eidheim: Yeah, at some point later we should add message boxes, for instance top left of the window, that disappears after a few seconds. There might be support for this in gtk already, will investigate in the summer.

eidheim:
Reversed iteration fixed, also now looks for non-class function implementations. I'll combine the last commit together with the first after you have looked through the changes.

zalox:
Looks good!

The commits are fine, they are descriptive and does what they tell us. I only want to avoid those "fixes to last commit.", because they interrupt workflow when iterating through changes. I think it's more important to maybe be more careful about pushes, and therefore consider doing a `git rebase -i` if there are git-cleanup needed since the last push. I don't want to promote `--force` pushing since that can be extremely dangerous.

eidheim:
Sounds good, merging then. Added smaller/temporary juCi++ messages (not in terminal) to my TODO.

PR #214, opened 2016-05-06
fixed issue with removing breakpoints

### LINKED ISSUE #212: Can't remove breakpoints
While the debugged program is running, breakpoints can't be removed (that is, juCi acknowledges the breakpoint removal by removing line highlighting, but the actual breakpoint isn't removed and LLDB stops on it anyway).

Quick debugging reveals that it's not removed because `breakpoint_path==file_path` is false at `debug_clang.cc:465` because `breakpoint_path` is `"/home/milleniumbug/dokumenty/PROJEKTY/NotMine/jucipp/build/debug/../../src/debug_clang.cc"` and `file_path` is `"/home/milleniumbug/dokumenty/PROJEKTY/NotMine/jucipp/src/debug_clang.cc"`

I assume using `boost::filesystem::equivalent` instead would fix the problem.

milleniumbug:
Fixes #212

### DIFF
--- a/src/debug_clang.cc
+++ b/src/debug_clang.cc
@@ -463,9 +463,10 @@ void Debug::Clang::remove_breakpoint(const boost::filesystem::path &file_path, i
             auto file_spec=line_entry.GetFileSpec();
             boost::filesystem::path breakpoint_path=file_spec.GetDirectory();
             breakpoint_path/=file_spec.GetFilename();
-            breakpoint_path = boost::filesystem::canonical(breakpoint_path);
+            boost::system::error_code ec;
+            breakpoint_path = boost::filesystem::canonical(breakpoint_path, ec);
             if(breakpoint_path==file_path) {
-              if(!target.BreakpointDelete(breakpoint.GetID()))
+              if(!ec || !target.BreakpointDelete(breakpoint.GetID()))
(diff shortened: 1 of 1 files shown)


src/debug_clang.cc
```diff
@@ -463,6 +463,7 @@ void Debug::Clang::remove_breakpoint(const boost::filesystem::path &file_path, i
             auto file_spec=line_entry.GetFileSpec();
             boost::filesystem::path breakpoint_path=file_spec.GetDirectory();
             breakpoint_path/=file_spec.GetFilename();
+            breakpoint_path = boost::filesystem::canonical(breakpoint_path);
```
eidheim: See for instance https://github.com/cppit/jucipp/blob/master/src/juci.cc#L29. Add `error_code ec` and check if `!ec`.

milleniumbug: And then...? What should be the appropriate response if the path handed out by LLDB doesn't exist? Is it because they got it wrong? Is it because someone removed it in the TOCTTU fashion? Should we crash? I suppose not. Should we ignore it? Sure, but then what is the state of the program? After all, the highlighting is removed by GUI somewhere else, so we can't just log it and move on. I ask because I seriously don't know.

eidheim: lldb is not 100% stable, at least not the earlier versions, it's just to be on the safe side. No need in my opinion to handle the error in this case, as it probably never will happen.

PR #215, opened 2016-05-09
Add draw_spaces option

hia3:
Determines what kind of whitespaces should be drawn. Use comma-separated
list of: space, tab, newline, nbsp, leading, text, trailing or all.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -186,6 +186,7 @@ void Config::get_source() {
   source.font=source_json.get<std::string>("font");
 
   source.cleanup_whitespace_characters=source_json.get<bool>("cleanup_whitespace_characters");
+  source.draw_spaces=source_json.get<std::string>("draw_spaces");
 
   source.show_map = source_json.get<bool>("show_map");
   source.map_font_size = source_json.get<std::string>("map_font_size");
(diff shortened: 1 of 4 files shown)


hia3:
1) With Boost 1.60 this will generate warnings (from Spirit), but this will be fixed (I hope) in 1.61 (almost released). Is it ok to use Spirit in this project at all?
2) Should I rename the option from "draw_spaces" to "show_spaces" as in "show_line_numbers"?
3) Why don't you use raw string literals in files.h?

eidheim:
Thank you, I'll have a look at this in a day or two. Default value should be empty I think (no whitespace shown). 
1) You might be able to disable the warnings by adding `system` in CMakeLists.txt like this:
`include_directories(SYSTEM ${global_includes})`
2) show_whitespaces? If newlines are shown, this is not a space but whitespace I guess
3) Actually, I have not tried raw string literals, but it seems like a good idea to use it (if g++4.8 supports it)

The currently oldest system we support now is Ubuntu 14, but in a few months we'll stop supporting Ubuntu 14. Then, the oldest supported system will be Debian stable. If this works on these two old systems I'm happy.

hia3:
I've set it to empty by default and fixed the warning. Let's continue when you have time.

eidheim:
Great, and also, before I forget, we have to up the JUCI_VERSION to ""1.1.3-2" so that people get the new preference item.

PR #217, opened 2016-05-10
Added brief info messages to top right corner

eidheim:
@zalox Feel free to merge if you are happy with it

* Animations (scroll down/up) will be added once https://bugzilla.gnome.org/show_bug.cgi?id=710888 is resolved. 
* Never mind the enormous diff in `src/source_clang.cc`, it's just that diff fails to handle indentation changes in a readable manner.

### DIFF
--- a/src/info.cc
+++ b/src/info.cc
@@ -1,0 +1,29 @@
+#include "info.h"
+
+Info::Info() {
+  set_hexpand(false);
+  set_halign(Gtk::Align::ALIGN_END);
+  
+  auto content_area=dynamic_cast<Gtk::Container*>(get_content_area());
+  label.set_max_width_chars(40);
+  label.set_line_wrap(true);
+  content_area->add(label);
+    
+  auto provider = Gtk::CssProvider::create();
+  provider->load_from_data("* {border-radius: 5px;}");
+  get_style_context()->add_provider(provider, GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
+}
+
+void Info::print(const std::string &text) {
+  if(!enabled)
+    return;
+  
+  timeout_connection.disconnect();
+  timeout_connection=Glib::signal_timeout().connect([this]() {
+    hide();
+    return false;
+  }, 3000);
+  
+  label.set_text(text);
+  show();
+}
--- a/src/info.cc
+++ b/src/info.cc
@@ -1,5 +1,19 @@
 #include "info.h"
 
+namespace sigc {
+#ifndef SIGC_FUNCTORS_DEDUCE_RESULT_TYPE_WITH_DECLTYPE
+  template <typename Functor>
+  struct functor_trait<Functor, false> {
+    typedef decltype (::sigc::mem_fun(std::declval<Functor&>(),
+                                      &Functor::operator())) _intermediate;
+    typedef typename _intermediate::result_type result_type;
+    typedef Functor functor_type;
+  };
+#else
+  SIGC_FUNCTORS_DEDUCE_RESULT_TYPE_WITH_DECLTYPE
+#endif
+}
+
 Info::Info() {
   set_hexpand(false);
   set_halign(Gtk::Align::ALIGN_END);
@@ -12,6 +26,15 @@ Info::Info() {
   auto provider = Gtk::CssProvider::create();
   provider->load_from_data("* {border-radius: 5px;}");
   get_style_context()->add_provider(provider, GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
+  
+  //Workaround from https://bugzilla.gnome.org/show_bug.cgi?id=710888
+  //Issue described at the same issue report
+  //TODO: remove later
+  auto revealer = gtk_widget_get_template_child (GTK_WIDGET (gobj()), GTK_TYPE_INFO_BAR, "revealer");
+  if (revealer) {
+    gtk_revealer_set_transition_type (GTK_REVEALER (revealer), GTK_REVEALER_TRANSITION_TYPE_NONE);
+    gtk_revealer_set_transition_duration (GTK_REVEALER (revealer), 0);
+  }
 }
 
 void Info::print(const std::string &text) {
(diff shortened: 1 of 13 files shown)


eidheim:
Tested on most platforms. Only Ubuntu 14 and debian stable has too old gtkmm versions to show the info messages.

src/source_clang.cc
```diff
@@ -859,20 +860,22 @@ Source::ClangViewAutocomplete(file_path, language) {
...
-        if(token.get_kind()==clang::Token_Identifier && cursor.has_type()) {
-          if(line==token.offsets.first.line-1 && index>=token.offsets.first.index-1 && index <=token.offsets.second.index-1) {
-            if(static_cast<unsigned>(token.get_cursor().get_kind())==103) //These cursors are buggy
-              continue;
-            auto referenced=cursor.get_referenced();
-            if(referenced)
-              return Token(this->language, static_cast<int>(referenced.get_kind()), token.get_spelling(), referenced.get_usr());
-          }
+    if(!parsed) {
+      Info::get().print("Buffer is parsing");
```
zalox: Maybe we should introduce some kind of assert function for this use case? I don't mind these, but it would look cleaner in my opinion.

src/source_clang.cc
```diff
@@ -916,22 +921,24 @@ Source::ClangViewAutocomplete(file_path, language) {
...
-              break;
-            }
+    if(!parsed) {
+      Info::get().print("Buffer is parsing");
+      return location;
+    }
+    auto iter=get_buffer()->get_insert()->get_iter();
+    auto line=(unsigned)iter.get_line();
+    auto index=(unsigned)iter.get_line_index();
+    for(auto &token: *clang_tokens) {
```
zalox: C-style cast.

src/info.h
```diff
@@ -0,0 +1,22 @@
+#ifndef JUCI_INFO_H_
+#define JUCI_INFO_H_
+
+#include <gtkmm.h>
+
+class Info : public Gtk::InfoBar {
+  Info();
+public:
```
zalox: Is Info() private? Move to private for clarity maybe?

src/info.cc
```diff
@@ -0,0 +1,29 @@
...
+  label.set_line_wrap(true);
+  content_area->add(label);
+    
+  auto provider = Gtk::CssProvider::create();
+  provider->load_from_data("* {border-radius: 5px;}");
+  get_style_context()->add_provider(provider, GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
+}
+
+void Info::print(const std::string &text) {
+  if(!enabled)
```
zalox:
```cpp
void Info::print(const std::string &text, unsigned timeout=Config::get().foo) {
```
I would like to be able to set timeout myself.

src/info.h
eidheim: Info is private here yeah, this is according to the C++ standard.

src/source_clang.cc
eidheim: This is not new code, but can change it anyway.

src/source_clang.cc
eidheim: I'm not entirely sure if I understand you here:) Just remember that I've not changed the refactoring methods, only some indentation mostly.

src/info.cc
eidheim: We should not populate the preference file with such never-going-to-be-used items I think.

src/info.h
eidheim: Also, constructors are always on start of class declaration.

src/source_clang.cc
zalox:
```cpp
if(!parsed) {
  Info::get().print("Buffer is parsing");
  return location
}
// could be something like
if(Info::get().assert(parsed, "Buffer is parsing"))
  return location;
```
I only saw this because it seems a condition is checked before every print and we get a lot of duplication. I realise it might be to much abstraction when thinking of readability :P

eidheim: As I understand if, assertion is typically used to identify things that never should happen and could lead to crashes. Using it here would be non-typical use.

src/source_clang.cc
eidheim: @zalox Fixed

src/info.cc
zalox: I disagree, when something is obstructing the field of view in the editor I think it should be configurable. We don't need to add the option to the config file, but users be should able to add the option if they want.

milleniumbug:
May I ask what are the guidelines on using this thing? On what basis should I decide whether to show information to the user by a terminal, or by the brief messages?

src/info.cc
eidheim: Not sure if timeout is the way to go then. Enable/disable maybe, but still, it gives information that maybe for instance plugins want to give, so disabling does not make sense in that respect. Let's discuss this later, and prioritise other things now:)

eidheim:
The idea is not to spam the terminal with non-essential smaller messages.

src/info.cc
eidheim: @zalox We could change the preferences like this maybe:
```json
{
    "version": "1.1.3-1",
    "gtk_theme": {
        "name_comment": "Use \"\" for default theme, At least these two exist on all systems: Adwaita, Raleigh",
        "name": "",
        "variant_comment": "Use \"\" for default variant, and \"dark\" for dark theme variant. Note that not all themes support dark variant, but for instance Adwaita does",
        "variant": ""
    },
    "window": {
        "default_size": {
            "width": "1024",
            "height": "768"
        },
        "show": {
          "directories": true,
          "terminal": true,
          "source_map": true,
          "info_messages": true,
          "info_messages_dealy": 3,
        }
    },
    "terminal": {
        "history_size": "1000",
        "font_comment": "Use \"\" to use source.font with slightly smaller size",
        "font": ""
    },
    ...
}
```
But let's consider this after this PR. 

At the moment, we can't have "hidden" preference items. Preference items not found in files.h will be removed when version gets bumped. This is needed for us to be able to change the preference item structure between versions, and users not to have unused nodes after upgrading juCi++. Adding "hidden" preference items would require us to add exceptions to this rule.

zalox: I'm going to merge this now, since it brings a very nice feature to juCi++. We can discuss configs later if we decide to refactor it :)

Issue #177, opened 2016-04-02
the IDE doesn't pick up standard settings and include directories


milleniumbug:
Opening up the project https://github.com/milleniumbug/juci_issue_reproducing_1/tree/4c44d5c9278b61662e6ef3f0fde725fb8a70480c in juCi++ shows errors (Clang seems to think it's in C++03 mode and it doesn't have any added include directories)

https://imgur.com/a/d4PAn

This is the content of the `compile_commands.json` file:

```
[
{
  "directory": "/home/milleniumbug/dokumenty/PROJEKTY/InDevelopment/sample_test/build",
  "command": "/usr/lib64/ccache/c++    -I../include   -std=c++11 -Wall   -o CMakeFiles/sample_test.dir/main.cpp.o -c /home/milleniumbug/dokumenty/PROJEKTY/InDevelopment/sample_test/main.cpp",
  "file": "/home/milleniumbug/dokumenty/PROJEKTY/InDevelopment/sample_test/main.cpp"
}
]
```

I'm on Fedora, running juCi++ cloned from commit 31343961dade904e941147a27bc7b2d485e144ba

eidheim:
Strange that your include (`-I../include`) is not an absolute path. What version of cmake do you have? I'll check other OS'es. 

Regarding the .hpp file, a temporary solution is to add an empty CMakeLists.txt in the include directory. Currently, juCi++ searches for these files to find the project root (the CMakeLists.txt with `project()` set). I'll see if I can fix that in some decent way. Maybe do an exception on header-files, as implementation files normally have a CMakeLists.txt in its path.

eidheim:
The problem goes away with header files and implementation files in the same directory, with a CMakeLists.txt in this directory.

milleniumbug:
`cmake --version` says this:

> `cmake version 3.4.1`

Speaking of errors I have similar issues when opening juCi++ itself in the IDE which I can't explain at all. Opening `source_clang.cc` is especially problematic.

https://imgur.com/a/nXEtn

To be honest, I wouldn't be surprised if clang would be totally broken on Fedora, I'll try on another system when I'll have the time.

zalox:
I sometimes get these full pages with red lines when there are errors related to preprocessor macros.

eidheim:
In general, I think its a good idea to use `${CMAKE_SOURCE_DIR}` when specifying project directories, for instance: 
```cmake
include_directories("${CMAKE_SOURCE_DIR}/include")
```
[...]

PR #218, opened 2016-05-11
Clicking on compiler error in terminal

hia3:
navigates to the file with error

### DIFF
--- a/src/terminal.cc
+++ b/src/terminal.cc
@@ -1,7 +1,8 @@
 #include "terminal.h"
-#include <iostream>
 #include "config.h"
 #include "project.h"
+#include "notebook.h"
+#include <iostream>
 
 Terminal::InProgress::InProgress(const std::string& start_msg): stop(false) {
   start(start_msg);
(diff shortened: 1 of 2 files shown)


eidheim:
Thank you again, I'll merge this Tomorrow morning most likely. Although I'll probably make some slight changes.

hia3:
How about using the same format (file:line:offset) in last_session.json/files? So may be you'll do something to reuse while doing your changes. :)

eidheim:
I would rather not create a 100% full path:line:index parser, especially if someone decided to put ':' and spaces in their file names. The current one works in 99% of the cases, and I'm happy with that.

eidheim:
Also, I'll add View::place_cursor_safe and other methods later to reduce the duplicate code found in refactoring methods and in Terminal::on_button_press_event.

milleniumbug:
Doesn't work when Ninja is used as a generator. Probably the same reason as [this](https://github.com/cppit/jucipp/issues/177).

eidheim:
I'll fix it Tomorrow morning. Thanks!

hia3:
I'd like to mention 2 minor things: 
1) errors from cmake (make?) could be made clickable (this is why I had to write that nested try block), marked on the picture;

not_highlighted

2) in last_session.json paths look weird:

    "files": [
        "D:\\dev\\jucipp\\src\\source.cc",
        "D:\/dev\\jucipp\\src\\terminal.cc"
    ],

Here source.cc was opened from the tree, but terminal.cc was opened from terminal link.

eidheim:
@hia3 Would anyone want to open the autogenerated build files that are marked in red? If so, I could mark them as well as links (and the fix to the issue @milleniumbug posted will fix opening these files too). 

Regarding 2, I'll have a look at the MSYS2 paths as well Tomorrow. My initial tests showed that it worked, but paths on MSYS2 are still sometimes problematic. Just to be sure, have you upgraded MSYS2 lately?

eidheim:
@hia3 I think the weird looking paths on Windows is only a visual issue, and does not affect juCi++ at all otherwise. Please feel free to correct me if I'm wrong.

eidheim:
See above commit. I decided not to add links to autogenerated build files, I think most people do not want to touch these!:)

PR #220, opened 2016-05-12
Continuous Integration with docker-compose 

zalox:
This PR is includes  #207.


eidheim:
Splendid! @milleniumbug you should probably have a look at this, as I currently only understand about 50% of this:)

eidheim:
Could we get error on warnings by adding the `-Werror` flag on Travis CI? I use the following on my own dockers:
```sh
cmake -DCMAKE_CXX_FLAGS="-Werror" -DCMAKE_CXX_COMPILER=g++ .. &&
make install
```

eidheim:
Additionally, after thinking about it, the most important distributions to test at the moment (before Ubuntu 16 gets older than debian stable in one year or so): 
  * debian stable (as the oldest distribution)
  * arch linux (as the newest distribution, but if there are too much trouble with testing on such a bleeding edge distribution, manjaro, being approximately one week behind, might be a good choice). 

Just in case running the tests on more than two distributions takes too long time. We should also compile with both gcc and clang too:) And do static analysis through http://clang-analyzer.llvm.org/scan-build.html.

zalox:
I'm adding osx support first, then I'll look at your options. I'll see what I can do throughout the night :)

zalox:
@eidheim I'm having trouble installing llvm with homebrew because of timeouts due to long compile time. Are there any ways of installing the binaries from http://llvm.org/releases/download.html ?

eidheim:
The binaries does not contain liblldb it seems, but you can use these instead of brew I think, as they are likely compiled with a approximately the same clang++ that is available through XCode (which brew uses). Just copy the prebuilt lib contents to /usr/local/lib, and maybe the rest of the prebuilt directories to /usr/local. It's fine to test the compilation without lldb I guess. If you want to include lldb, one would have to create a llvm brew package with clang and lldb I guess that one could tap or something. I do not know how to do this, however, one might base it on the current llvm brew package. 

But, can you install OS X on a docker container?? Which image are you using:)?

zalox:
There isn't an osx container, but travis provides a virtual machine, but I decided to drop osx for now, since I shouldn't be the one doing it, I understand now there is more to homebrew than I thought.

.travis.yml
```diff
@@ -9,6 +9,11 @@ env:
   - distribution: debian-testing
   - distribution: fedora
   - distribution: arch
+  - distribution: ubuntu-scan-build
```
eidheim: You can probably to the scan-builds in the above distributions, no need for running compilation both with and without scan build I think. Also exit with error if the html directory gets created, or maybe there is a scan-build option to report error if issues are found?

eidheim:
@zalox Regarding scan-build, I think, or am pretty sure, that one should instead run this in a separate docker image, that is instead of in each of the docker images. The reason is that on has to run the following to make it work:
```sh
...
mkdir jucipp/build
cd jucipp/build
scan-build cmake ..
scan-build --status-bugs make
```

You can use the arch linux image for this I guess, to get the newest llvm static analyzer.

The output of scan-build cmake .. is as follows, and note the c*-analyzer as compiler:
```sh
scan-build: Using '/usr/bin/clang-3.8' for static analysis
-- The C compiler identification is GNU 6.1.1
-- The CXX compiler identification is GNU 6.1.1
-- Check for working C compiler: /usr/lib/clang/ccc-analyzer
-- Check for working C compiler: /usr/lib/clang/ccc-analyzer -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/lib/clang/c++-analyzer
-- Check for working CXX compiler: /usr/lib/clang/c++-analyzer -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found LibClang: /usr/lib64/libclang.so  
-- Boost version: 1.60.0
-- Found the following Boost libraries:
--   regex
--   system
--   filesystem
-- Found ASPELL: /usr/lib64/libaspell.so  
-- Found PkgConfig: /usr/bin/pkg-config (found version "0.29.1") 
-- Checking for module 'gtkmm-3.0'
--   Found gtkmm-3.0, version 3.20.1
-- Checking for module 'gtksourceviewmm-3.0'
--   Found gtksourceviewmm-3.0, version 3.18.0
-- Could NOT find plantuml (missing:  PLANTUML_JARFILE) 
-- Could NOT find Doxygen (missing:  DOXYGEN_EXECUTABLE) 
-- Configuring done
-- Generating done
-- Build files have been written to: /home/eidheim/jucipp/build
scan-build: Removing directory '/tmp/scan-build-2016-05-23-085035-29239-1' because it contains no reports.
scan-build: No bugs found.
```

PR #221, opened 2016-05-14
feature: show warning when files have changed on disk outside of juCi++

zalox:
Shows info message when a file has changed on disk.


eidheim:
Looks good! I guess its ok that the info message shows several times on the same file when cycling through the tabs for instance. 

Also, not related to this PR, but I'll look into a way to calculate the delay of the info messages based on the number of characters. Maybe there is some commonly used algorithm for this.

PR #229, opened 2016-05-22
Added initial tests

eidheim:
To enable testing ~~(only works with g++, clang did not support the needed linking options)~~:
```sh
cmake -DENABLE_TESTING=1 ..
make
make test
```

If you are on a terminal only environment without a display server, like Travis CI:
```sh
broadwayd&
make test
```

### DIFF
--- a/tests/cmake_test.cc
+++ b/tests/cmake_test.cc
@@ -4,9 +4,11 @@
 #include <boost/filesystem.hpp>
 
 int main() {
-  CMake cmake(JUCI_TESTS_PATH);
+  auto tests_path=boost::filesystem::canonical(JUCI_TESTS_PATH);
   
-  g_assert(cmake.project_path.filename()=="jucipp");
+  CMake cmake(tests_path);
+  
+  g_assert(cmake.project_path==boost::filesystem::canonical(tests_path/".."));
   
   auto functions_parameters=cmake.get_functions_parameters("project");
   g_assert(functions_parameters.at(0).second.at(0)=="juci");
--- a/tests/cmake_test.cc
+++ b/tests/cmake_test.cc
@@ -13,8 +13,8 @@ int main() {
   auto functions_parameters=cmake.get_functions_parameters("project");
   g_assert(functions_parameters.at(0).second.at(0)=="juci");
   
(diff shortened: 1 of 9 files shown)


eidheim:
@zalox have a look at this. Tested on most platforms.

tests/cmake_test.cc
```diff
@@ -0,0 +1,18 @@
+#include <glib.h>
+#include "cmake.h"
+#include "project_build.h"
+#include <boost/filesystem.hpp>
+
+int main() {
+  CMake cmake(JUCI_TESTS_PATH);
+  
+  g_assert(cmake.project_path.filename()=="jucipp");
+
```
zalox: Test fails if parent directory is named `"juci"` for instance.

eidheim: Well, don't rename it;)? It's just a sample test though, can probably be improved. This PR is mostly a proof of concept, and directions and samples for creating future tests.

eidheim: Could change it to `!cmake.project_path.empty()` though

zalox: I don't want to force the users to name their source directory jucipp either. :P

eidheim: Good point

eidheim: Fixed!

eidheim:
@zalox I can `git rebase` this if you'd like after you have said it looks good for merge. Added some extra cmake tests.

zalox:
https://travis-ci.org/cppit/jucipp/builds/132117723 Can you look at this? The clang 3.8 seems to fail on warnings.

eidheim:
Actually, it might be that "note: remove std::move call here" is correct. This is great!:)

zalox:
@eidheim I know! That's why I merged it prematurely ;)

eidheim:
Good job!

eidheim:
@zalox https://news.ycombinator.com/item?id=9915300:
"clang 3.7 has a warning when std::move prevents it from using return value optimization"
You can remove the `std::move`'s in `src/filesystem.cc` I think:) Although, feel free to read up more on this, but my initial though of course is that clang is correct.

I got to read student reports now!

eidheim:
Also, I saw you had the tests run too! Well done!

eidheim:
Finally, I think you can add an image link to https://travis-ci.org/cppit/jucipp.svg?branch=master in README.md too?

PR #233, opened 2016-05-25
CMakeLists.txt and test cleanup

eidheim:
* Added source_clang_test to figure out how the build should look like
* In order to avoid duplicate compilation of files, project_shared_sources library was added that is used in both ./src and ./tests. This drastically improves the compilation speed of the tests.
* General cleanup of CMakeLists.txt's to avoid duplicate code in src/CMakeLists.txt and tests/CMakeLists.txt


eidheim:
@zalox could you have a look at this? We should merge this soon so I can continue to make tests. 

By the way, CI is awesome!:)

eidheim:
@zalox Do you know why Travis CI was not run on the last commit above?

zalox:
I see no reason for why the CI didn't build. The only thing I've noticed is that it won't run if there are syntax errors in `.travis.yml`.

I'm reviewing now :)

tests/source_clang_test.cc
```diff
@@ -0,0 +1,43 @@
...
+//broadwayd&
+//make test
+
+int main() {
+  auto app=Gtk::Application::create();
+  Gsv::init();
+  
+  Config::get().project.default_build_path="./build";
+  Source::ClangView *clang_view=new Source::ClangView(boost::filesystem::canonical(std::string(JUCI_TESTS_PATH)+"/clang_project/main.cpp"),
+                                                      Gsv::LanguageManager::get_default()->get_language("cpp"));
```
zalox: Raw pointer for fast destruction?

eidheim: Yeah the destruction for Source::ClangView is a bit special due to libclang.

src/source_clang.cc
```diff
@@ -1112,18 +1112,16 @@ Source::ClangViewAutocomplete(file_path, language) {
...
-    auto iter=get_iter_for_dialog();
-    selection_dialog=std::unique_ptr<SelectionDialog>(new SelectionDialog(*this, get_buffer()->create_mark(iter), true, true));
-    auto rows=std::make_shared<std::unordered_map<std::string, clang::Offset> >();
-    auto methods=clang_tokens->get_cxx_methods();
-    if(methods.size()==0)
-      return;
-    for(auto &method: methods) {
+    auto cxx_methods=clang_tokens->get_cxx_methods();
+    if(cxx_methods.size()==0)
+      return methods;
```
zalox: I think this check is redundant, the for-loop below does essentially the same thing. I realise this wasn't added by this commit though.

zalox: I just realised it became redundant in this commit.

tests/source_clang_test.cc
eidheim: You are right, I'll fix it later.

src/source_clang.cc
eidheim: You are right, the if statement should be removed.

eidheim: Fixed!

PR #237, opened 2016-06-01
CI with appveyor on windows with msys2

zalox:
* CI on windows with msys2
* More work on osx, but disabled since I can't manage to install llvm with clang


zalox:
@eidheim I'm merging this if the build succeeds.

Issue #235, opened 2016-05-30
window split (feature request)


milleniumbug:
![current](http://i.imgur.com/IyfIWYC.png)

I'd like to have two visible files at the same time because a.) often I want to see two files at the same time (either because I want to see header file and implementation file at the same time, or I want to look up a thing in a file, but I don't want to hide existing file because it's distracting, lots of reasons) b.) the amount of free space here is staggering.

Sadly, opening two instances of juCi++ is not a suitable replacement.

eidheim:
Agree! Have thought about this many times, but have always dropped it due to the added complexity, especially when the parsing of a tab is triggered by saving another tab. But I think it's doable anyway. My suggestion is as follows when split window is triggered from menu:
* Add a notebook to the right with its own tabs (no tabs initially)
* Make it possible to drag and drop tabs between the two notebooks
* Open a new file in the notebook that currently has focus
* For simplicity, it should not be possible to have the same file opened in two notebooks
* Start reparsing, if needed, if the notebook tab gets focus, that is do not immediately start reparse a file even though it is shown on screen. 
* Current class Notebook will be renamed Notebooks, and some methods will be extended to support two notebooks. Most likely, the changes will only happen in this class. Notebooks::get_current_view for instance will return the view/tab with focus.
* `~/.juci/last_session.json` has to store information about which notebook the different files are opened in. 

I do not have time for this now, and I'd like to start working on git integration soon, but I think split window can be implemented during the summer. 

However, it would be great help if someone could create an as-simple-as-possible-one-file-example of two notebooks where it is possible to drag and drop tabs from one notebook to another.

eidheim:
Actually, according to http://www.kksou.com/php-gtk2/sample-codes/move-tabs-between-two-notebooks-with-drag-and-drop.php and https://developer.gnome.org/gtkmm/stable/classGtk_1_1Notebook.html#a7b0555d623c2784dc45b5b3e6aa9c3e1, this seems much easier than I initially though. I'll fix it in a week or two.

eidheim:
@milleniumbug see the PR above.

PR #239, opened 2016-06-02
Added split window

eidheim:
I think this is stable now, but I guess we should test it a week or so just to be sure. I made some choices to avoid too much complexity, but feel free to give me feedback. 

See also #235.


eidheim:
@zalox could you look at the cases when travis failed above?

milleniumbug:
This is awesome! There's only a minor issue: width of the split windows isn't remembered across restarts, which I'm fine with, but I'd expect a sane default to be provided, like 50:50 or 60:40, not this:

disproportionate windows

eidheim:
See Notebook::toggle_split, I think the positioning is done correctly. What gtk version do you have? Drag and drop between notebooks is broken in debian testing at the moment for instance, and positioning works perfect on OS X and pretty well on Manjaro. There might be some bug reports on gtk on this I maybe that might shed some light on the issues.

eidheim:
@milleniumbug Found this regarding the positioning: https://bugzilla.gnome.org/show_bug.cgi?id=765676. Would you mind test if the workaround works?

milleniumbug:
Replacing `set_position(get_width()/2);` with `int width = get_width(); set_position(width/2-1); set_position(width/2);` did nothing.

Tried debugging it, and the value of `width` at this point is 479. Now I suspect this is `i3-wm` specific issue (the window manager I'm using)

zalox:
@eidheim You should pull in master into this branch, so we get CI with windows as well :)

zalox:
@eidheim Regarding the failed builds, it seems travis fails due to a docker/ubuntu bug, and windows seems to fail due to resource management in appveyor. I think you can restart builds if any of these scenarios happen, if not I shall give you the missing permissions.

eidheim:
@milleniumbug split window position should be fixed in https://github.com/cppit/jucipp/pull/239/commits/0e0ef9ba19321853aa2b7119726a16d93debb27d (last commit above). At least it's correct now in Manjaro.

eidheim:
@zalox this PR is getting ready for merge. Maybe a day or two more with testing, and after @milleniumbug has verified that the split window position is correct now.

milleniumbug:
Can confirm that now after restart waits for a moment and then sets the divider 50:50.

eidheim:
@zalox this PR seems quite stable. Feel free to merge after review.

eidheim:
Merged, since new important PR is coming.

PR #249, opened 2016-07-04
Fix #240 / extended fix to MSYS2's libclang to '-isystem'

### LINKED ISSUE #240: code parser does not honour cmake_prefix_path?
I pass the command -DCMAKE_PREFIX_PATH=... to cmake (configured in the config.json) as I have some libraries in a non standard path. In the CMakeLists.txt I also do "include_directories(...)".

Building of the project works as expected, but the parsing of the code shows lots of errors, as the include-files are not found - only include-files in the standard paths are found..

donbcd:


eidheim:
Thank you, looks good!

PR #257, opened 2016-07-12
Replaced NULL with nullptr wherever applicable.

nshcat:
I did the dirty work and replaced all occurrences of C-Style `NULL` macro usage in API calls with C++11's `nullptr` as they are guaranteed to be compatible.


eidheim:
Thank you! Merging soon.

PR #262, opened 2016-07-22
GUI changes in preparation for additional debug information: terminal…

eidheim:
… is now directly below notebook

@zalox ready for merge if you agree on the gui changes, and after you have tested the changes. 

The result is shown below, and the red circle shows where additional debug info like watchpoints and current frame variables will be positioned when its implemented:
image


zalox:
Hi! Sorry for the late response on this! I agree on these changes.

PR #263, opened 2016-07-23
added git status command to menu

Toeger:
Added the menu entry Source->Git->Status. It saves unsaved files if `save_on_compile_or_run` is true and it calls `git status` in the project directory.
The output of git status is printed to the terminal. It may need to be made prettier.
The shortcut (probably `CTRL+G+S`) is missing, couldn't figure out how to do it.

### DIFF
--- a/src/menu.cc
+++ b/src/menu.cc
@@ -132,6 +132,10 @@ Menu::Menu() {
         <submenu>
           <attribute name='label' translatable='yes'>_Git</attribute>
           <item>
+            <attribute name='label' translatable='yes'>_Status</attribute>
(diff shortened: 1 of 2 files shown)


eidheim:
Thank you for looking into this, but with juCi++ we are not trying to replace using your OS and/or a real tty-terminal. The various git commands can just as well be done in a terminal, or even using the Run Command menu item. Some of the output of `git status` can also be seen by the file/directory colours in the directory view.

zalox:
Hi, thank you for your interest in juCi++! I have to agree with eidheim. juCi++ shouldn't have too many third-party builtins, so I'm closing this. I'm looking forward to your next suggestion :)

PR #270, opened 2016-08-31
Introduced a method to clean build directory

d3rrial:
**Do not actually merge this, please.**
I wrote an e-mail to @eidheim earlier and proposed a method to clean up the build directory.
Now this is obviously a hideous proof of concept and _whether or not such a cleanup function is actually useful_ is very debatable. There are no mutexes set or anything, this is not write-proof, I deleted dirty with boost, etc. as I said, proof of concept, please do not merge!

I'm just putting this here, just in case.

PS: Sorry for the whitespace mess, I used juCi++ to edit it and still had "clean up whitespace" enabled in the options and that turned into what you see now.


eidheim:
Thank you! Currently, the way to achieve this is to remove the build directory by hand, and then reopen the files. Another option, to rebuild with the cmake cache files, is to save a CMakeLists.txt in the project directory. One could also remove the build directory and then save a opened CMakeLists.txt and thus avoid reopening the already opened c/c++ files. 

On the other hand, it might maybe be useful, especially for less experienced cmake users, to have an option to do a complete rebuild (deleting the build directory and rebuilding it) without having to reopen or saving any files. I'll add this to my TODO list, and if you could do a commit without removing our beloved whitespaces, I could continue on your commit. In this way, you get credited for this work as well. 

Regarding the whitespaces, we basically just keep them around so that we don't have to be too strict when accepting commits, and also we then never have to have meaningless commits that would just remove any whitespaces we would maybe miss. Although, the commits should be readable:)

eidheim:
I did a git rebase in my repository, and thus this PR did not get automatically marked as merged. It is merged however, thank you very much:)

PR #272, opened 2016-08-31
Fixes segmentation fault when using ctags on very long lines

zalox:
A segmentation fault occurs when `Ctags::get_location` gets a large line. 

I don't know whether this is normal behaviour, but when juCi++ fell back to ctags it parsed the `./docs` directory as well. A html file had a line of > 40 000 chars, resulting in a failure by the regex library.


zalox:
I chose 2048 since some compilers consider that maximum method signature length.

eidheim:
Great, merging now.

PR #273, opened 2016-09-01
Added way to clear console manually

d3rrial:
Enables the user to clear the console manually by selecting Window -> Clear Console ~~or hitting Ctrl+Shift+C (keyboard shortcut pending review)~~ edit: or by setting a new keyboard shortcut

### DIFF
--- a/src/files.h
+++ b/src/files.h
@@ -141,7 +141,7 @@ R"RAW(
 R"RAW(
         "close_tab": "<primary>w",
         "window_toggle_split": "",
-        "clear_console": "<primary><shift>c"
+        "window_clear_terminal": ""
     },
     "project": {
         "default_build_path_comment": "Use <project_directory_name> to insert the project top level directory name",
--- a/src/files.h
+++ b/src/files.h
@@ -140,7 +140,8 @@ R"RAW(
 #endif
 R"RAW(
         "close_tab": "<primary>w",
-        "window_toggle_split": ""
+        "window_toggle_split": "",
+        "clear_console": "<primary><shift>c"
     },
     "project": {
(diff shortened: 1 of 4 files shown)


d3rrial:
Sorry that I'm messing with your IDE so much... I'm sure you'd rather have your peace and quiet

eidheim:
Thank you! And contributions are very welcome! I'll look at this later Today or Tomorrow morning.

src/files.h
```diff
@@ -140,7 +140,8 @@ R"RAW(
 #endif
 R"RAW(
         "close_tab": "<primary>w",
-        "window_toggle_split": ""
+        "window_toggle_split": "",
+        "clear_console": "<primary><shift>c"
```
eidheim: Please use window_clear_console. The default key could be "" in case we want to use control-shift-c for something else later. Or?

eidheim: Also increase the minute version number in CMakeLists.txt when adding another configuration item.

eidheim: When increasing the version number, an older preference file gets automatically updated. That is, when someone downloads the latest changes and there are new preference items available, or old ones have been removed (or renamed).

eidheim: Maybe window_clear_terminal? Since we use terminal in the source code, and elsewhere.

d3rrial: Alright, I'll change it

src/menu.cc
```diff
@@ -352,7 +352,7 @@ Menu::Menu() {
       <section>
         <item>
           <attribute name='label' translatable='yes'>_Clear _Console</attribute>
```
d3rrial: Argh, forgot to change the text there, one more commit, sorry

eidheim:
Thank you again, merging very soon.

PR #274, opened 2016-09-01
Corrected install instructions for OS X

Istarnion:
gtksourceviewmm3 is now properly part of homebrew and can be installed directly instead of using homebrew/x11/gtksourceviewmm3.


eidheim:
Thank you! Merging now.

Issue #278, opened 2016-09-03
Breakpoint Line-Tag (Red Circle)


d3rrial:
Is it in scope to implement such a marker in juCi++? All the big IDEs I've used before have that (Such as Visual Studio or CLion), the question is if you even want it.

I've done a clumsy example implementation here: https://github.com/d3rrial/jucipp/commit/69f09613c802ee8e44517747a31e0dc3a584d8f9 But that only displays the dot where the git diff tags go, so it doesn't look very pretty:
![breakpoint_dot_cropped](https://cloud.githubusercontent.com/assets/3718129/18224687/afd98d06-71de-11e6-8bed-82bfcdaec1fe.png)

And it also doesn't activate on click, instead just with the normal Ctrl+B breakpoint shortcut.

eidheim:
Are not these marker a remnant from the old days when you did not have the possibility to colour lines? Also use the keyboard instead of the mouse;)

d3rrial:
I don't know. Personally I just like the markers visually.

And I also think it's a waste to have a mouse connected to your PC if you don't use it ;) I always hear about workflow interruptions, but I personally work quite well even with the mouse in my workflow...

As for where it comes from, no idea. I've suggested it because it might be a familiar feature for people who switch from other IDEs to juCi++. Whether it's useful or not, I'm not sure. I personally like being able to click on a line number and have it become a breakpoint and the dot where you clicked is a good visual indication that it actually worked, too.

eidheim:
One possibility could be to add a breakpoint on the line when one clicks on the line number, but without the red circle. You are right, there is the use case when one use the mouse to scroll the source code, and one wants to add a breakpoint without using the keyboard to do so.

d3rrial:
I'm not trying to push for immediate implementation btw. Maybe just put this suggestion somewhere on a "To Consider" list. It's not an important feature after all. Ctrl+B is quick to press and the red line background is a good indicator.

eidheim:
No problem, but I think adding the breakpoint with mouse click is a good solution, if you're interested in looking at the implementation for this. That is, without the red circle:)

d3rrial:
I will look into it. I've not completely understood how the GTK Gutter works, so it'll take a few days or so. If I end up unable to find out how to do it, I'll put another comment on this issue.

eidheim:
Great, thanks alot again!

PR #279, opened 2016-09-03
Toggle breakpoint on click in line-number area (Gutter)

d3rrial:
Added rudimentary functionality to toggle breakpoints on and off by mouseclick into the righthand side of the gutter (to the right of the line numbers).

It's not exactly pretty yet and I required to include project.h and notebook.h into the source_diff.cc which is probably not as you intend. The toggle debug is also (almost) a duplicate of the one used in project.cc, so maybe this can be changed.

I talked about this in this issue: https://github.com/cppit/jucipp/issues/278

Note: I received a lot of help while implementing this.


src/source_diff.cc
```diff
@@ -54,6 +75,11 @@ Source::DiffView::DiffView(const boost::filesystem::path &file_path) : Gsv::View
   renderer->tag_removed_below=get_buffer()->create_tag();
   renderer->tag_removed_above=get_buffer()->create_tag();
   
+  add_events(Gdk::BUTTON_PRESS_MASK | Gdk::BUTTON_RELEASE_MASK);
+  signal_line_mark_activated().connect(sigc::mem_fun(*this, &Source::DiffView::gutter_click_handler));
+  
+  set_show_line_marks(true);
```
d3rrial: This should probably be handled somewhere else.

src/source_diff.cc
```diff
@@ -54,6 +75,11 @@ Source::DiffView::DiffView(const boost::filesystem::path &file_path) : Gsv::View
   renderer->tag_removed_below=get_buffer()->create_tag();
   renderer->tag_removed_above=get_buffer()->create_tag();
   
+  add_events(Gdk::BUTTON_PRESS_MASK | Gdk::BUTTON_RELEASE_MASK);
```
d3rrial: I don't know for what events you're already checking, so I added this to be sure, but it's probably redundant

d3rrial:
Strange... Why do the tests fail? It built fine for me... Let me do a clean rebuild

d3rrial:
OK, this clearly warrants a lot more work. I don't really know where to stick the handler code, but it doesn't seem right where it's currently at.

Where are the views instantiated? This should probably be where I put the code, but I couldn't find it (yet)

eidheim:
The views are instantiated in notebook.cc. Don't worry too much about where to place it, if you have the functionality, I can do the cleanup job:) Great work so far, have not studied it yet, but hopefully I'll have a look Tomorrow. Just a quick question, is it possible to add/remove the breakpoints by clicking on the line numbers instead? Just thinking that we should keep the gutter as slim as possible, in case you have made it broader?

d3rrial:
I wish it was, but it's not, unfortunately. That part of the gtk gutter does not have any eventhandlers. I could shoehorn something crazy with using the x & y coordinates to figure out the location, but that wouldn't cover the line_nr so I'd have to figure out a solution for that.
GTK just doesn't come equipped for that, so it'd be a terrible hackjob, if it worked at all.

eidheim:
I tried https://developer.gnome.org/gtksourceviewmm/stable/classGsv_1_1Gutter.html#a74ad7512e18f6a5a2ccb6cdca160bdf3 (Gsv::Gutter::get_renderer_at_pos), but I have not yet gotten it to work. It should though, and then one could use the https://developer.gnome.org/gtksourceviewmm/stable/classGsv_1_1GutterRenderer.html#a9a4adadc30f1c2dd9b198c442be2b153 (Gsv::GutterRenderer::signal_activate). 

I saw one implementation of Gsv::Gutter::get_renderer_at_pos, where only the x value is used. The implementation is probably like this still. According to https://developer.gnome.org/gtksourceview/stable/GtkSourceGutter.html, the line numbers should be at position -30.

eidheim:
Got it working:
```diff
diff --git a/src/source_diff.cc b/src/source_diff.cc
index 9e743b8..47b3b39 100644
--- a/src/source_diff.cc
+++ b/src/source_diff.cc
@@ -76,11 +76,31 @@ Source::DiffView::DiffView(const boost::filesystem::path &file_path) : Gsv::View
   renderer->tag_removed_above=get_buffer()->create_tag();
   
   add_events(Gdk::BUTTON_PRESS_MASK | Gdk::BUTTON_RELEASE_MASK);
-  signal_line_mark_activated().connect(sigc::mem_fun(*this, &Source::DiffView::gutter_click_handler));
-  
-  set_show_line_marks(true);
   
   configure();
+  
+  signal_realize().connect([this] {
+    auto gutter=get_gutter(Gtk::TextWindowType::TEXT_WINDOW_LEFT);
+    auto render=gutter->get_renderer_at_pos(30, 0);
+    if(render)
+      render->signal_activate().connect([this](const Gtk::TextIter& iter, const Gdk::Rectangle&, GdkEvent*) {
+        if(auto view=Notebook::get().get_current_view()) {
+          auto line_nr=iter.get_line();
+          if(view->get_source_buffer()->get_source_marks_at_line(line_nr, "debug_breakpoint").size()>0) {
+            auto start_iter=view->get_buffer()->get_iter_at_line(line_nr);
+            auto end_iter=view->get_iter_at_line_end(line_nr);
+            view->get_source_buffer()->remove_source_marks(start_iter, end_iter, "debug_breakpoint");
+            if(Project::current && Project::debugging)
+              Project::current->debug_remove_breakpoint(view->file_path, line_nr+1, view->get_buffer()->get_line_count()+1);
+          }
+          else {
+            view->get_source_buffer()->create_source_mark("debug_breakpoint", view->get_buffer()->get_iter_at_line(line_nr));
+            if(Project::current && Project::debugging)
+              Project::current->debug_add_breakpoint(view->file_path, line_nr+1);
+          }
+        }
+      });
+  });
 }
 
 Source::DiffView::~DiffView() {
```

eidheim:
Now, after moving the changes from source_diff to source.cc, and there create a public method called something like `void toggle_breakpoint(int line_nr)` should make this PR mostly complete. Move the toggle breakpoint code to this function, and call this function in some constructor as done in the diff above in source.cc. Also call this function in the toggle_breakpoint action in window.cc.

eidheim:
But then including project.h will become necessary in source.cc. I'll see if there is a way around that later.

eidheim:
At least we get rid of the notebook.h include.

PR #280, opened 2016-09-04
Added breakpoint on click functionality

d3rrial:
Same as https://github.com/cppit/jucipp/pull/279 but without the detour.

Will still fail the tests for some reason, most likely the Project singleton.

### DIFF
--- a/src/source.cc
+++ b/src/source.cc
@@ -4,6 +4,7 @@
 #include "terminal.h"
 #include "info.h"
 #include "directories.h"
+#include "project.h"
 #include <gtksourceview/gtksource.h>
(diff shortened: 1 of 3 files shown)


eidheim:
Don't worry, I'll fix it. Thank you again!

d3rrial:
Well, you did most of the work, turned out that it was possible after all with the line numbers and I just looked in the wrong place... >_>

eidheim:
Gtkmm is quite nice to work with once it is somewhat understood, before that however it can be pretty hard. 

I've done the fixes, so will merge Today I think, just need to do some testing.

PR #282, opened 2016-09-04
cleanup: make global includes global

zalox:


src/CMakeLists.txt
```diff
@@ -2,18 +2,6 @@ if(MSYS)
...
-set(global_includes
-   ${Boost_INCLUDE_DIRS}
-   ${GTKMM_INCLUDE_DIRS}
-   ${GTKSVMM_INCLUDE_DIRS}
-   ${LIBCLANG_INCLUDE_DIRS}
-   ${ASPELL_INCLUDE_DIR}
-   ${LIBGIT2_INCLUDE_DIRS}
-   ../libclangmm/src
-   ../tiny-process-library
-   .
```
eidheim: Think it should be `../src` instead of `.` to make the tests compile. And juci will compile as well too then I guess.

CMakeLists.txt
```diff
@@ -68,6 +68,18 @@ set(global_libraries ${global_libraries}
...
+set(global_includes
+   ${Boost_INCLUDE_DIRS}
+   ${GTKMM_INCLUDE_DIRS}
+   ${GTKSVMM_INCLUDE_DIRS}
+   ${LIBCLANG_INCLUDE_DIRS}
+   ${ASPELL_INCLUDE_DIR}
+   ${LIBGIT2_INCLUDE_DIRS}
+   ../libclangmm/src
+   ../tiny-process-library
+   .
```
eidheim: Think it should be `../src` instead of `.` to make the tests compile. And juci will compile as well too then I guess.

zalox: @eidheim See 425f8ef.

tests/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
...
-    stubs/selectiondialog.cc
-    stubs/terminal.cc
-    stubs/tooltips.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/stubs/config.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/stubs/dialogs.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/stubs/directories.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/stubs/info.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/stubs/selectiondialog.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/stubs/terminal.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/stubs/tooltips.cc
```
eidheim: Great, but just a minor comment: `${CMAKE_CURRENT_SOURCE_DIR}` is used here, but not in: https://github.com/cppit/jucipp/pull/282/commits/425f8ef52baaf89e142c6b9cf7796da1266ea723#diff-95e351a3805a1dafa85bf20b81d086e6L33. I guess `stubs/config.cc` here for instance is ok instead of `${CMAKE_CURRENT_SOURCE_DIR}/stubs/config.cc`, or what do you think?

zalox: > cmake docs:
```cmake
# Relative paths are interpreted as relative to the current source directory.
```
It seems they are redundant :) I'll remove them.

zalox:
@eidheim Ready for merge I think.

PR #283, opened 2016-09-05
Show file path in the info bar at the bottom

d3rrial:
I haven't created functionality to disable this from the configuration, so this should probably go on a ToDo list...

I added this because I think it's helpful to know which folder I'm working in. I have jucipp cloned 4 times on my drive, one that's just the standard cppit/jucipp clone and 3 working copies (partially from my fork) and it isn't immediately apparant which one I'm working on, unless I go into the "open folder" dialog or search the files for changes I've made.

The path it shows is always the one the cursor is currently in, same as the line-column indicator next to it. If I've understood your code correctly, the path label should only be updated when the view is changed, but I'm not sure, please look over it if possible :)

This is what it looks like:
file_path_cropped

Edit:
Tested with split mode and in debug mode with a very small window size. The debug status will be pushed off-center by this, but it remains readable. And split mode works just as intended, the path of the file which has the cursor is shown.

### DIFF
--- a/src/notebook.cc
+++ b/src/notebook.cc
@@ -178,6 +178,10 @@ void Notebook::open(const boost::filesystem::path &file_path, size_t notebook_in
     if(get_current_view()==view)
       info.set_text(" "+info_text);
   };
+  source_views.back()->on_update_path_label=[this](Source::View* view, const std::string &path_label_text) {
+    if(get_current_view()==view)
+      path_label.set_text(" "+path_label_text+" ");
+  };
   
   scrolled_windows.emplace_back(new Gtk::ScrolledWindow());
   hboxes.emplace_back(new Gtk::HBox());
--- a/src/notebook.h
+++ b/src/notebook.h
@@ -44,6 +44,7 @@ public:
 
   Gtk::Label info;
   Gtk::Label status;
+  Gtk::Label path_label;
   
   std::function<void(Source::View*)> on_change_page;
   std::function<void(Source::View*)> on_close_page;
--- a/src/source.cc
+++ b/src/source.cc
@@ -863,6 +863,11 @@ void Source::View::set_info(const std::string &info) {
     on_update_info(this, positions+" "+info);
 }
 
+void Source::View::set_path_label(const std::string &path_label){
+  if(on_update_path_label)
+    on_update_path_label(this, path_label);
+}
+
 std::string Source::View::get_line(const Gtk::TextIter &iter) {
   auto line_start_it = get_buffer()->get_iter_at_line(iter.get_line());
   auto line_end_it = get_iter_at_line_end(iter.get_line());
--- a/src/source.h
+++ b/src/source.h
@@ -86,8 +86,10 @@ namespace Source {
     
     std::function<void(View* view, const std::string &status_text)> on_update_status;
     std::function<void(View* view, const std::string &info_text)> on_update_info;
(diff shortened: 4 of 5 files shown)


d3rrial:
I don't know how to code for your git integration, so see this as a feature request:
Could you also include something like the path-thing I've added here, but instead of showing the path, show which branch is currently checked out in git? I'd pack it at the end instead of the start tho.

eidheim:
Great, I'll look at it in a day or so! @zalox are you happy with adding the file name on the bottom?

d3rrial:
Also getting these warnings sometimes:

```
(juci:26231): Gtk-WARNING **: Allocating size to gtkmm__GtkWindow 0x19cd170 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?

(juci:26231): Gtk-WARNING **: Allocating size to gtkmm__GtkWindow 0x19ccc70 without calling gtk_widget_get_preferred_width/height(). How does the code know the size to allocate?
```
I'm not 100% sure, but I think they appear when the autocomplete dialog opens.

eidheim:
That warning has been there a while on some systems. Not sure if it's gtk or something we do in selectiondialog.cc. Feel free to look into it:) It's unrelated to this PR though.

d3rrial:
@eidheim unrelated because I posted in the wrong window... \*facepalm\* I'll look into it regardless ;)

eidheim:
I'm thinking, keeping things minimalistic, that the full file path can be part of the existing label that shows line:column. Also, the branch can be in parentheses after the file name, for instance:
`/home/jucipp/some_repository/main.cpp(master):4:2`
We use filename:line:column elsewhere in juCi++ as well. Thoughts?

d3rrial:
I don't think it's good that way, because usually when you want to see the line:column label, you'll want to know the information on it at a single glance, so a fixed position is best for that. The file path has a variable length however, so the line:column marker would move with the length of the file path.

It sounds a bit weird when I say it like that, because you know that the L:C marker is at the end of the path, but to find the end of the path the eyes still have to do a little bit of scanning to find it, whereas with the marker at the beginning of the line your eyes can just directly go to the bottom left and receive the required information.

The difference is minimal I think, but my gut feeling says the L:C marker in a fixed position and the path info floating behind it

eidheim:
I basically agree with you, but the warnings and errors status would move as well unless we make a fixed space for it somewhere. And it is not very often one looks at the line and column number anymore, with all the helps IDE's are giving the user nowadays.

d3rrial:
Do what you think is best. I honestly don't know which is better. I like the way you integrated the current branch tho. :+1:

eidheim:
Alright, I'll give it some more thought, and see if I can get the branch name in there somehow, and also see if I can find a fixed position for the warnings and errors. I probably need a week or so though, and I will probably make a new PR on this. Hope that is ok with you.

d3rrial:
Sure! Thank you!

zalox:
I'm quite happy with this :+1:  Haven't been able to test it yet though.

eidheim:
@d3rrial I added some commits to my branch at https://github.com/eidheim/jucipp regarding the changes discussed above. Have not yet added branch info, but when I do I'll create a new PR. Feel free to try it out, and let me know what you think:)

eidheim:
Due to the complexity of this work, I replaced this PR with PR #288

Issue #281, opened 2016-09-04
Autogenerated closing brackets


d3rrial:
Currently when autocompleting a function/method in juCi++, opening and closing brackets are generated automatically, along with sample content. In other IDEs (Visual Studio / CLion) the automatically generated parentheses are what I'll call unsolid, meaning if `test(/* ... */)` is generated and my cursor is here: `test(/* ... */ꕯ)`, and I hit the closing bracket on my keyboard ")", the new closing bracket I wrote is ignored / replaces the auto-generated bracket, which allows me to then finish the line with ;↵ and be on my merry way with typing in the next line.

In juCi++ however the autogenerated closing bracket is solid, so when my cursor is right before the bracket and I hit a closing bracket, a new closing bracket is generated and I now have two. To get around this I either have to hit the delete key to delete the autogenerated bracket, or never manually close the bracket and skip the created one with the right arrow key, both of which force me to leave the regular working set of keys and enter the numpad area.

I don't know where you handle autogenerated function calls, if you point me in a direction, maybe I can work something out (no promises tho).

I don't know how much this issue affects others, but I personally find it a jarring break in typeflow.

eidheim:
You can probably fix this in `Source::View::on_key_press_event_basic`

d3rrial:
I'll look into it, thanks

d3rrial:
OK, negative from me, I'm unable to figure out. I don't really understand your code well enough to do this.
I'll leave the issue open as a feature request, alright?

eidheim:
Could you write down in pseudocode the wanted behaviour maybe? And which characters should be considered (for instance `)`, `]` and `}`).

eidheim:
I ask mainly in case there are any exceptions to the rule you described above.

eidheim:
One example exception could be if any text is selected.

d3rrial:
Can't write it in pseudocode because I don't know how I would check for bracket-consistency.
What I mean is something like this:

- User accepts auto-complete dialog for a function call
- user inputs function parameters
- when user is before the closing ")" of the generated function call and hits ")" to insert the closing bracket which had already been generated, the input is ignored and the cursor is moved behind the generated ")".

Kinda like this.
I haven't found the autocompletion to be disturbing in any other occurrence.

How do you mean if text is selected?

eidheim:
[...]

PR #284, opened 2016-09-05
Indentation cleanup, added smart_brackets and smart_insertions, and more

eidheim:
@d3rrial, @zalox and others: I need testing of this PR before merge:
* Added move one ) instead of insert, which is very nice after autocomplete (#281)
* Major cleanup of the indenting code, it's actually much faster compared to current master, and additionally more readable (I hope). 
* smart_insertions
* Various other improvements


eidheim:
And the indentation is improved as well. For instance it should now correctly indent:
```c++
some_function(true, 42,
              [] {//enter here
```

, that will lead to:

```c++
some_function(true, 42,
              [] {
  //cursor here
}
```

d3rrial:
```
void asdf(std::string& superduperlongvariablenamewtf, std::string& andanotheronewhowouldbelievethis,
          std::string& thisiswaytoomuch){
            
          }
```
Is the indent here as intended? Looks strange to me.

Other than that it looks good to me.

eidheim:
@d3rrial Thank you, that is definitely a bug. I'll fix it Tomorrow.

eidheim:
I fixed the bug @d3rrial mentioned. Also improved indentation when writing constructor implementations on multiple lines. Let me know if there are other issues:

eidheim:
See also the last. You can now avoid using arrow keys when for instance writing:
```c++
auto c=std::make_shared<int>(10);
```

src/source.h
```diff
@@ -124,9 +124,10 @@ namespace Source {
     
     Gtk::TextIter find_start_of_sentence(Gtk::TextIter iter);
     bool find_open_non_curly_bracket_backward(Gtk::TextIter iter, Gtk::TextIter &found_iter);
+    bool find_close_non_curly_bracket_forward(Gtk::TextIter iter, Gtk::TextIter &found_iter);
     bool find_open_curly_bracket_backward(Gtk::TextIter iter, Gtk::TextIter &found_iter);
     bool find_close_curly_bracket_forward(Gtk::TextIter iter, Gtk::TextIter &found_iter);
-    long open_close_paranthesis_count(Gtk::TextIter iter);
+    long open_close_bracket_count(Gtk::TextIter iter, unsigned int left_bracket, unsigned int right_bracket);
```
d3rrial: `unsigned int`? Are you sure? Wouldn't that mean the `'('` has to be cast from char? Or does Clang handle single quoted symbols differently?

eidheim: `*iter` returns `unsigned int` since it is a UTF-8 character. So I do the cast automatically in the function parameters. At least the tests passed:)

d3rrial:
`auto c=std::make_shared<int>(10);`
Works well, but not flawlessly (this might be out of scope for this PR, tho):

when I type auto `c=std::make_sha` I get an autocomplete suggestion for

`auto c=std::make_shared<typename _Tp>(_Args &&__args...)`

where my cursor sits in the angled bracketes with `typename_Tp` being highlighted (so it's overwritten when I start typing). To get past that initial `>` after the `typename_Tp` I still have to use the arrow key to skip it, tho. Other than that, it works perfectly.

src/source.h
d3rrial: Well it wasn't an error, at most I would have expected the compiler to throw a warning. As long as it works, it's all good :)

eidheim:
In the case of `auto c=std::make_shared<typename _Tp>(_Args &&__args...)` and you replace typename _Tp you should get moved when entering `>` just before the `>`. You should be able to enter both template argument and function argument without using the keys after using autocompletion on std::make_shared.

d3rrial:
This doesn't work for me. Could it be because of different keyboard layouts / keyval differences?

eidheim:
Yeah, tested on Linux. Have to find the right key that works (GDK_KEY_greater works on OS X).

d3rrial:
I just tested and GDK_Key_greater is keycode 62, but 62 is also they keyval for > for me, so that's probably not the cause...

eidheim:
The problem is searching for `<` and `>` keys it seems. I'll see if I can fix it later.

d3rrial:
Well, `<` and `>` are also comparison operators after all. In the case where it's used as a templating mechanism, I think it's enough to stop as soon as any other type of bracket is found. I can't think of a situation where `<...)>` or `<...]>` or `<...}>` would be valid constructs.

```cpp
auto a=crazytemplate<std::string, void*, int,
                     double, size_t>(...);
```
This has to remain valid, tho...

d3rrial:
Looks good to me, even works with nested templates. Haven't done any edge-case tests for false positives yet, but I haven't had any false negatives. :+1:

eidheim:
Thank you, good to hear. At least there is some progress:)

eidheim:
I made the rules for the movements very strict so that they will not happen unless you want them to, hopefully.

d3rrial:
While I personally love this feature / these movements and will/would use them all the time, for some people this might be an inconvenience (I can't fathom how, because they behave just like regular, solid brackets in any case other than when you type in a bracket which is already there) so I think it might be best if this was put into a setting in the config file.

eidheim:
I'm thinking that too. I'll use it myself, but we should maybe have it disabled by default. What do you think @zalox ?

zalox:
See  https://github.com/cppit/jucipp/issues/281#issuecomment-244985254.

I see no reason to not enable them by default. It would effectively improve the coding experience for new users. People like me, that don't like these features, know how to disable them and keep quiet as long as there is an option to do so.

eidheim:
So what should we call this bracket feature in the preferences? `smart_brackets`? I'll add some comment as well (`smart_brackets_comment`).

eidheim:
Not completely sure about the preference item name `smart_brackets`, but at least the functionality is there.

eidheim:
I went a bit crazy and implemented "smart insertations". Needs testing!:)

eidheim:
Added some improvements, and decided to drop smart insertations on `{}`; it was just too annoying at least the way I work:)

d3rrial:
Works perfectly in the few scenarios I came up with. I'm going to test this more long term to be sure, but from my perspective this looks good!

zalox:
I'm having a look tonight :) @tedjk @oyvang You should also test juCi++ now.

eidheim:
It's not 100% perfect yet, and one important rule I'll add is that one should never add pairs when in front of a letter, underscore or a number. For instance:
`if(|true) //cursor at |`
should become
`if((|true)` instead of `if((|)true)`

There are probably more such rules as well. 

There is also some minor issue when the cursor is in front of `"something"` or `'s'`.

eidheim:
Also, I guess that it does not make sense to add `()` in front of a `(`. The result should be `((...)` instead of `()(...)` I take it?

d3rrial:
Well, unless you want to C-cast the result of a calculation

eidheim:
Good point!

eidheim:
But still I think it is more annoying that `()` comes up when you want `(`, than the other way around. Though, this is a personal preference. We should try keep the false positives to a minimum though I belive. But would love to hear your opinions on this.

d3rrial:
I totally agree with you on that. C-style casts shouldn't be used in C++ anymore anyway, so this is not relevant very often and not having the bracket appear does not cause any inconvenience I believe.

Btw. do you have a keyboard shortcut to surround a given highlighted text with brackets?

eidheim:
I know gtk has some such functions that one can create shortcuts for (in a gtk-init file), but have not looked into this much.

eidheim:
A large cleanup again, also activated smart brackets/insertions for all files, and added exceptions for () and [] insertions.

eidheim:
Fixed various smaller issues in last commit.

d3rrial:
I've found one thing that I dislike:
If I type
```cpp
void a
```
and open the bracket, it autocompletes to:
```cpp
void a()
```
if I (for some reason) don't want the second bracket tho and hit "Delete" to delete the character to the right of the cursor, both brackets get deleted and I'm back with
```cpp
void a
```
but unable to open just a single bracket. This is resolved by typing a single letter within the brackets, but personally I'd prefer it if I could hit the delete key and delete only characters to the right of the cursor.

eidheim:
@d3rrial fixed in last commit, thank you!

eidheim:
Added some more fixes, getting closer to the finishing line:) We should probably test this for some extra days though. Let me know if there are more issues:)

eidheim:
Ok, I'm happy with this PR, at least right this minute. It should be pretty optimised too. We should test it some more days though. And if there is anything that is annoying, even the slightest, let me know:)

d3rrial:
I'm a little sad that this doesn't seem to work with other languages (for example Rust) for which I'm also using juCi++. Maybe allow specifically smart `""` and `()` insertion in other languages as well? I can't think of a high level language where the insertion rules for strings or brackets would be different from the syntax C++ uses. (But my knowledge of languages is also quite limited)

eidheim:
It should be enabled for all languages, but it depends on how good the language definition files are for gtksourceview. You can use it in CMakeLists.txt for instance, but cmake's language definition files are missing a string definition (it still works though).

d3rrial:
Oh, OK. Strange. I would've thought it would work then... Rust has complete syntax highlighting and keyword autocomplete. I must've overestimated how good the language definition files are, then. Thanks!

eidheim:
See the language definitions in `/usr/share/gtksourceview-3.0/language-specs/`. For instance python.lang has <context ... class=...>, but the rust.lang file is missing this attributes. You can add them your self though. juCi++ looks for class or class-disabled attributes in the context nodes, and if found enables for instance smart_insertions to work. See https://github.com/cppit/jucipp/blob/master/src/source.cc#L1853.

d3rrial:
Well, since it's obviously working from your side but not from theirs, I'll just wait for better Rust integration into gtksourceview. Thanks again! :)

eidheim:
I think I fixed it. It was basically my bad.

d3rrial:
And now I sound passive aggressive... :P
Thanks for the fix :)

eidheim:
Just added smart insertions on selected text.

d3rrial:
I like it.
I'm still internally debating if it would be better in its own setting or if it's good like it is. On one hand it's really useful when you want to surround something with brackets, on the other hand it overrides the common notion that typing a character while having something highlighted will replace the highlighted text.

I'm personally for leaving it as it is.
I think it when the time comes for documention this should be properly explained, tho.

eidheim:
We have some freedom when a user activates this option I think. Also added insertions of
```
`
```
(dead grave) for markdown on selected text.

eidheim:
I'm going to rename `smart_brackets` to `smart_moves` I think (in preferences and elsewhere). Does that make sense? Maybe rename `smart_insertions` to `smart_inserts` too? Not sure about the latter one though.

d3rrial:
I think `smart_inserts` is better than `smart_insertions`, but I like `smart_brackets` more than `smart_moves`
smart_moves seems so nondescript if you don't know the context.

eidheim:
Let's just keep it as it is then, easiest solution anyway since I can't decide:) AutoPairs does not include inserting `;` for instance, so it is not descriptive enough I think.

eidheim:
I also did a spellcheck cleanup now that we have a working is_code_iter! The spellcheck code should be more understandable now.

eidheim:
Ah, and testing on the new spellcheck code is needed as well:)

d3rrial:
That reminds me, is there a way to disable the spell-check entirely? It often flags my variable names for silly reasons. It's not that important since it has no influence on my workflow, it's just a visual annoyance.

eidheim:
You can set the spellcheck language to some invalid value in preferences. However, variable names outside strings and comments should not be spellchecked.

d3rrial:
Since the latest commit or before that already?
I think I had the spellchecker tell me off for variable names, but it could've also been strings, I'm tired right now and I can't trust my memories. I'll test it again tomorrow and report back.

eidheim:
The spellchecker should be context aware, and has been since the beginning, but there are some occasions where it fails to underline properly like when doing undo/redo/pasting, but it should not underline code at all. If it does, it is removed after a short delay.

eidheim:
Fixed the spellcheck issue on undo/redo/pasting.

zalox:
Shouldn't the `{` also be smart_inserted?

eidheim:
I tried doing smart insertion of {}, but found that too often that was not what I wanted. Additionally, we already have smart indentation and insertion of } after pressing enter after {, and this leads to far less false positives.

eidheim:
@zalox This PR should be getting ready for merge. I added some more improvements that should be pretty safe. The biggest change is the indentation cleanup, and that seems to work fine.

eidheim:
@zalox this PR should now be ready for merge!:)

eidheim:
I'll merge this now as this PR is stable and contains numerous improvements.

Issue #285, opened 2016-09-09
Reload files changed outside of juCi++


d3rrial:
When a file is changed outside of juCi++ the only way I found to reload the file was to close its tab in juCi++ and double click it in the tree view. Is it possible to add a menu-point / key-combo to reload the document that currently has the cursor?

If there already is one, I couldn't find it, if so, is it possible to make it more visible / obvious? Like for example putting a clickable label in the notification:

> Caution: juci.cc was altered outside of juCi++ __Click here to reload it__

d3rrial:
This is what I came up with:
https://github.com/d3rrial/jucipp/commit/a1a0d9270f021b384eebb950908483b12f01ce48
I don't know if
```cpp
Notebook::get().close_current();
```
is the right way to go and it probably needs to make sure that a warning flashes if you have unsaved changes. Some other things probably too but I can't think of them

eidheim:
Thank you again, but sadly I'm quite busy till Tuesday and will not be able to look at new features till then (only have time for potential bugfixes).

d3rrial:
Sure. It's not urgent ;)

PR #287, opened 2016-09-13
Reload file

d3rrial:
Adds a menu option to reload the current file.
Normally when a file is changed from outside of juCi++, it first has to be closed completely and then re-opened in order to work with these changes.

In case of a race while using this PR (a file was changed externally and juCi++ has unsaved changes), upon clicking "Reload File" in the menu, the function will always be executed, but the user is asked if they want to save the current file before it's opened. There's no merging and no backups, either the file on the drive is overwritten or the changes within the editor are overwritten.

To fix this, it would make sense to add an extra warning beforehand, or, instead of an overwriting save, open the "Save File As" prompt, to give the user the option to save his current changes to a separate file and manually merge them afterwards.

I don't know how to do that, so it's not in this pull request.

This pull request addresses this issue I opened earlier: https://github.com/cppit/jucipp/issues/285

I'm going to try to find out how to solve the race-condition, but I'd like your input on that.

### DIFF
--- a/src/menu.cc
+++ b/src/menu.cc
@@ -63,6 +63,12 @@ Menu::Menu() {
       </section>
       <section>
         <item>
+          <attribute name='label' translatable='yes'>_Reload _File</attribute>
+          <attribute name='action'>app.reload_file</attribute>
+        </item>
+      </section>
+      <section>
+        <item>
           <attribute name='label' translatable='yes'>_Save</attribute>
           <attribute name='action'>app.save</attribute>
         </item>
--- a/src/notebook.cc
+++ b/src/notebook.cc
@@ -128,6 +128,38 @@ std::vector<Source::View*> &Notebook::get_views() {
   return source_views;
 }
 
+void Notebook::reload(const boost::filesystem::path &file_path, size_t notebook_index) {
+  if(boost::filesystem::exists(file_path)) {
+    std::ifstream can_read(file_path.string());
+    if(!can_read) {
+      Terminal::get().print("Error: could not open "+file_path.string()+"\n", true);
+      return;
+    }
+    can_read.close();
+  }
+  
+  auto last_view=get_current_view();
+  int offset = last_view->get_buffer()->get_insert()->get_iter().get_offset();
+  int line = last_view->get_buffer()->get_insert()->get_iter().get_line();
+
+  auto language=Source::guess_language(file_path);
+  last_view->get_buffer()->erase(last_view->get_buffer()->begin(), last_view->get_buffer()->end());
+  last_view->get_source_buffer()->begin_not_undoable_action();
+  if(language) {
+    if(filesystem::read_non_utf8(file_path, last_view->get_buffer())==-1)
+      Terminal::get().print("Warning: "+file_path.string()+" is not a valid UTF-8 file. Saving might corrupt the file.\n");
+  }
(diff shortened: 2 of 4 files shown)


eidheim:
Thank you, this is a good idea in case one is working on multiple branches. I think a yes/no dialog should be enough though on the reload action. Also the directories has to be updated (Directory::get().update()) on this action, and the source_diff's configure has to be run I think (without having looked at the source code). There should maybe be a update method in source_diff if there is no already.

d3rrial:
What do they do?
Because as it is right now it really just does reload the current file from the directory and treats it as a newly opened file (so it's parsed again)

eidheim:
See my updated comment:)

By the way, no need to close and reopen the tab, you can just clear the buffer and insert the contents of the new file. One should also try to go the the last cursor position if it's still there, if not somewhere nearby, and do a scroll to that position.

d3rrial:
I went with the path of least resistance here. I think that clearing/replacing the buffer with the contents would require new safety checks by hand. I'll look into it, but I can make no promises for success :)

d3rrial:
I've experimented quite a bit and I can't get it to work properly. Whenever I try to create a new textiter in the new buffer to place the cursor, the application instead marks the text from offset 0 to the cursor offset and if I manually place the cursor outside of the previously marked text, the whole application crashes with segfault and error code 139

I'm definitely doing something wrong, but I can't figure out what.

Same goes if I only replace the current current buffer with the new buffer and try to save (segfault, return code 139)

The "this file was changed outside of juCi++" warning also doesn't go away.

eidheim:
I don't have time to look at this right now, but you can use this new function I added some weeks back: https://github.com/eidheim/jucipp/blob/master/src/source.h#L51. It handles the gtk specifics related to placing the cursor somewhere nearby its previous position after buffer update.

d3rrial:
Ahh, I see now why it's crashing >_>

I'm letting an object run out of scope but I'm still referencing a deallocated pointer out of scope...

Can't believe I didn't spot that yesterday...
The buffer is a 
```cpp
Glib::RefPtr<Gtk::TextBuffer>
```
But the parent of this buffer is running out of scope. Any idea how I can let the parent run out of scope but preserve the buffer itself?

eidheim:
There should be no need to create a new Source::View since the file should have the same extension. Deleting a view is somewhat complex since it has to be done in the background due to there are no way of stopping libclang while it's processing. I think this will solve your issues.

d3rrial:
OK, added a new commit that does what you described and works without any crashes.
There are the following issues, which I can only resolve if I work on the newest build (probably) which I am currently not (because of difficulties with forking on github):
The "File was altered outside of juCi++" warning remains after having reloaded the file. I'll need it to reset the last write time, so that this warning goes away.
Reloading the file can't be undone with Ctrl+Z. I've done that on purpose here: ```last_view->get_source_buffer()->begin_not_undoable_action();```, because ```last_view->get_buffer()->erase(last_view->get_buffer()->begin(), last_view->get_buffer()->end());``` is seperate on the undo stack from loading the buffer, which means that hitting undo would first make the editor completely empty and only after a second undo would it restore to the original state. I don't know how to solve that.

eidheim:
Thank you! I did some changes, but it's merged now:)

PR #288, opened 2016-09-15
Status bar

eidheim:
* Added filename and branch to status bar at bottom
* Added colors to diagnostics; number of warnings, errors and fix-its
* Cleanup of status processing

Need feedback on the changes before merge.

### DIFF
--- a/src/config.cc
+++ b/src/config.cc
@@ -12,12 +12,12 @@ Config::Config() {
   for (auto &variable : environment_variables) {
     ptr=std::getenv(variable.c_str());
     if (ptr!=nullptr && boost::filesystem::exists(ptr)) {
-      home /= ptr;
-      home /= ".juci";
+      home_path = ptr;
+      home_juci_path = home_path / ".juci";
       break;
     }
   }
-  if(home.empty()) {
+  if(home_juci_path.empty()) {
     std::string searched_envs = "[";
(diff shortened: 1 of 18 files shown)


eidheim:
Have to mention that the decisions of the placements of the various items in the status bar was made to make the items take up minimal area, and avoiding that the items move around. The items should not overlap even when debugging.

d3rrial:
Works perfectly for me and looks good. I'm still not a fan of the line:col indicator at the back, but it also makes sense to be there (I mostly dislike it for visual reasons).
I have not been able to test it on windows, but it works great on Arch Linux. I especially like the branch indicator and that it changes immediately when a new branch is checked out.

I'll test on Windows in a few hours. Thank you! :)

eidheim:
Great:) I agree with the line:line_offset, though the alternatives are either to have the file name move around, or take up extra area (where the filename would move around if one is working on files with a large number of lines/line offsets). I think the current solution is the best, though none of alternatives are perfect. 

I was surprised by the branch indicator myself, and that the diff on the views are getting updated when one switch branches:) Also added a warning when switching branches (if the current file is different) to make this solution part of a normal work flow. The final step, is your PR where one can reload the file:)

eidheim:
Just made the diagnostics status an overlay too, so that it never moves around.

PR #290, opened 2016-09-20
Automatic include guards

sv1990:
Adds include guards automatically to files ending on .h, .hh, .hpp or .hxx.
When a file with the right extension is opened, e.g. foo.hh, it will open the file as
```
#ifndef _FOO_HH_
#define _FOO_HH_

|

#endif // _FOO_HH_
```
with the cursor placed at '|'. I included this as an option to preferences.


ff2000:
Besides pragma once being the way better solution (and also supported on non-MS-platforms/compilers):
The implementation is quite risky, it can mess up builds because the guard is easily predictable.
The guard should protect ;) I would suggest to include the project name, complete path and the unix timestamp
of the creation time. But I am sure there are people that have problems with such long guards, so probably
make it configurable, probably  with some predefined naming schemes?

On Tue, 20 Sep 2016 02:15:22 -0700, sv1990 <notifications@github.com> wrote:
> Adds include guards automatically to files ending on .h, .hh, .hpp or .hxx.
> When a file with the right extension is opened, e.g. foo.hh, it will open the file as
> ```
> #ifndef _FOO_HH_
> #define _FOO_HH_
> 
> |
> 
> #endif // _FOO_HH_
> ```
> with the cursor placed at '|'. I included this as an option to preferences. 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/cppit/jucipp/pull/290
> 
> -- Commit Summary --
> 
>   * Automatically set include guards depending on file extension
>   * Bugfix: Don't take substr on empty string
>   * Set cursor inbetween include guards
>   * Reduce code duplication
>   * Add automatic_include_guards as an options
> 
> -- File Changes --
> 
>     M src/config.cc (1)
>     M src/config.h (3)
>     M src/directories.cc (23)
>     M src/files.h (1)
> 
> -- Patch Links --
> 
> https://github.com/cppit/jucipp/pull/290.patch
> https://github.com/cppit/jucipp/pull/290.diff
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/cppit/jucipp/pull/290
> 
> 
--

sv1990:
Thanks for the feedback. You are right. I think i will implement the naming schemes
none
_PROJECT_FOO_HH_
_PROJECT_PATH_FOO_HH
_PROJECT_PATH_FOO_HH_TIMESTAMP_
so that one has increasing protection :)

eidheim:
Thank you, but it seems `#pragma once` has been supported by gcc and clang for a while (at least according to https://en.wikipedia.org/wiki/Pragma_once). I still see some manual include guards out there though, and we use them as well, but is there a good reason why we are not using `#pragma once` instead?

sv1990:
You are right. I was just used to using include guards and in emacs I just had to use a snippet to get them. 
`#pragma once` is probably the better way of doing it.

One other thing: If I use `#pragma once` in juci++ i get a warning "#pragma once in main file"

eidheim:
Aha, I'll see if I can get rid of that warning later. Might have to send `-Wno-pragma-once-outside-header` or something similar to libclang when working with header files in `source_clang.cc`.

eidheim:
@sv1990 I just pushed the pragma once warning fix to master.

sv1990:
@eidheim  Thanks

PR #291, opened 2016-09-22
feature: change font of status to monospaced

zalox:
In my opinion this looks better with a monospaced font on the status bar. Any objections?


eidheim:
I think that the main issue is that monospace fonts take up more space than non-monospace fonts. And we do not have much space in the status bar, especially for longer file names. Also, it is not a problem that different letters take up different area in the status bar, unlike the terminal and the source view.

PR #298, opened 2016-11-12
Added support for the Meson Build System

eidheim:
Should be 100%, but some additional testing needed. Let me know if there are any issues or not.


zalox:
I'll have a look, please give me some time to learn meson.

eidheim:
Sure, and thank you:)

src/compile_commands.cc
```diff
@@ -0,0 +1,79 @@
...
+  bool found_argument=false;
+  for(auto &parameter: parameters) {
+    if(found_argument) {
+      parameter_values.emplace_back(parameter);
+      found_argument=false;
+    }
+    else if(parameter==paramter_name)
+      found_argument=true;
+  }
+
```
zalox:
```c++
std::copy_if(parameters.begin(),paramerters.end(),parameter_values.end(),[](){}));
```

src/compile_commands.cc
```diff
@@ -0,0 +1,79 @@
...
+  }
+  
+  return parameter_values;
+}
+
+CompileCommands::CompileCommands(const boost::filesystem::path &build_path) {
+  try {
+    boost::property_tree::ptree root_pt;
+    boost::property_tree::json_parser::read_json((build_path/"compile_commands.json").string(), root_pt);
+
```
zalox: Correct me if I'm wrong, but I think read_json can cause problems. I think there is a error_code version.

src/compile_commands.cc
```diff
@@ -0,0 +1,79 @@
...
+        parameters.emplace_back(parameter);
+      };
+      for(size_t c=0;c<parameters_str.size();++c) {
+        if(backslash)
+          backslash=false;
+        else if(parameters_str[c]=='\\')
+          backslash=true;
+        else if((parameters_str[c]==' ' || parameters_str[c]=='\t') && !backslash && !single_quote && !double_quote) {
+          if(parameter_start_pos!=static_cast<size_t>(-1)) {
+            add_parameter();
```
zalox: std::npos?

src/meson.cc
```diff
@@ -0,0 +1,130 @@
...
+    return false;
+  if(!boost::filesystem::exists(default_build_path)) {
+    boost::system::error_code ec;
+    boost::filesystem::create_directories(default_build_path, ec);
+    if(ec) {
+      Terminal::get().print("Error: could not create "+default_build_path.string()+": "+ec.message()+"\n", true);
+      return false;
+    }
+  }
+
```
zalox: Suggestion:
```c++
if(project_path.empty() || default_build_path.empty())
    return false;
if(!boost::filesystem::exists(project_path/"meson.build"))
   return false;
```

src/meson.cc
```diff
@@ -0,0 +1,130 @@
...
+    return false;
+  if(!boost::filesystem::exists(debug_build_path)) {
+    boost::system::error_code ec;
+    boost::filesystem::create_directories(debug_build_path, ec);
+    if(ec) {
+      Terminal::get().print("Error: could not create "+debug_build_path.string()+": "+ec.message()+"\n", true);
+      return false;
+    }
+  }
+
```
zalox: This calls for abstraction, but I agree to wait if we are introducing more build systems.

src/compile_commands.cc
eidheim: The function gets more complicated if copy_if is to be used:
```c++
std::vector<std::string> CompileCommands::Command::paramter_values(const std::string &parameter_name) const {
  std::vector<std::string> parameter_values(parameters);
  
  bool found_argument=false;
  auto it=std::copy_if(parameters.begin(), parameters.end(), parameter_values.begin(),
               [&parameter_name, &found_argument](const std::string &parameter){
    if(found_argument) {
      found_argument=false;
      return true;
    }
    else if(parameter==parameter_name) {
      found_argument=true;
      return false;
    }
    return false;
  });
  parameter_values.resize(std::distance(parameter_values.begin(), it));
  
  return parameter_values;
}
```

src/meson.cc
eidheim: Will clean up this as well.

src/compile_commands.cc
zalox: std::copy_if(parameters.begin(),paramerters.end(),parameter_values.end(),[&parameter_name]const std::string &param(){
  return parameter_name == param;
}));

eidheim: This is not the same. The idea is to return for instance `some_file` after `-o some_file` when the `parameter_name` is `-o`.

zalox: Ohh, sorry I missed the else if, but at least this made me notice the miss spelt `paramter_name`.

eidheim: I fixed the misspelling yesterday:)

eidheim:
@zalox Are you happy with the changes? Can we merge this PR?

eidheim:
@zalox the spelling errors should be fixed now

zalox:
Looks good, sorry about my late responses, been flying all over Europe!

PR #302, opened 2016-12-05
New right click menu

zalox:
* Right clicking in source buffer now displays the edit buffer
* Right click selects the hovering line
* Able to toggle comments easy
* Rename rename to refactor rename

### DIFF
--- a/src/menu.cc
+++ b/src/menu.cc
@@ -3,9 +3,85 @@
 #include <string>
 #include <iostream>
 
-Menu::Menu() {
-  ui_xml = R"RAW(
-<interface>
+const Glib::ustring menu_xml= R"RAW(<interface>
+  <menu id='right-click-line-menu'>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Undo</attribute>
+        <attribute name='action'>app.edit_undo</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Redo</attribute>
+        <attribute name='action'>app.edit_redo</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Paste</attribute>
+        <attribute name='action'>app.edit_paste</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Reload _File</attribute>
+        <attribute name='action'>app.reload_file</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Go _to _Declaration</attribute>
+        <attribute name='action'>app.source_goto_declaration</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Go _to _Method</attribute>
+        <attribute name='action'>app.source_goto_method</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Go _to _Usage</attribute>
+        <attribute name='action'>app.source_goto_usage</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Rename</attribute>
+        <attribute name='action'>app.source_rename</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Toggle _Line _Comment</attribute>
+        <attribute name='action'>app.source_comments_toggle</attribute>
+      </item>
+    </section>
+  </menu>
+  <menu id='right-click-selected-menu'>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Undo</attribute>
+        <attribute name='action'>app.edit_undo</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Redo</attribute>
+        <attribute name='action'>app.edit_redo</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Cut</attribute>
+        <attribute name='action'>app.edit_cut</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Copy</attribute>
+        <attribute name='action'>app.edit_copy</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Paste</attribute>
+        <attribute name='action'>app.edit_paste</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Toggle _Line _Comment</attribute>
+        <attribute name='action'>app.source_comments_toggle</attribute>
+      </item>
+    </section>
+  </menu>
   <menu id='juci-menu'>
     <section>
       <item>
@@ -378,7 +454,6 @@ Menu::Menu() {
   </menu>
 </interface>
 )RAW";
-}
 
 void Menu::add_action(const std::string &name, std::function<void()> action) {
   auto g_application=g_application_get_default();
@@ -400,14 +475,18 @@ void Menu::set_keys() {
 }
 
 void Menu::build() {
-  builder = Gtk::Builder::create();
-  
   try {
-    builder->add_from_string(ui_xml);
-    auto object = Menu::get().builder->get_object("juci-menu");
+    builder = Gtk::Builder::create_from_string(menu_xml);
+    auto object = builder->get_object("juci-menu");
     juci_menu = Glib::RefPtr<Gio::Menu>::cast_dynamic(object);
-    object = Menu::get().builder->get_object("window-menu");
+    object = builder->get_object("window-menu");
     window_menu = Glib::RefPtr<Gio::Menu>::cast_dynamic(object);
+    object = builder->get_object("right-click-line-menu");
+    auto ptr = Glib::RefPtr<Gio::Menu>::cast_dynamic(object);
+    right_click_line_menu = std::make_unique<Gtk::Menu>(ptr);
+    object = builder->get_object("right-click-selected-menu");
+    ptr = Glib::RefPtr<Gio::Menu>::cast_dynamic(object);
+    right_click_selected_menu = std::make_unique<Gtk::Menu>(ptr);
   }
   catch (const Glib::Error &ex) {
     std::cerr << "building menu failed: " << ex.what();
--- a/src/source.cc
+++ b/src/source.cc
@@ -4,6 +4,7 @@
 #include "terminal.h"
 #include "info.h"
 #include "directories.h"
+#include "menu.h"
 #include <gtksourceview/gtksource.h>
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/spirit/home/qi/char.hpp>
@@ -2123,6 +2124,20 @@ bool Source::View::on_button_press_event(GdkEventButton *event) {
     return true;
   }
   
+  if((event->type == GDK_BUTTON_PRESS) && (event->button == 3)){
+    if(!get_buffer()->get_has_selection()){
+      int x,y;
+      window_to_buffer_coords(Gtk::TextWindowType::TEXT_WINDOW_TEXT,event->x,event->y,x,y);
+      Gtk::TextIter iter;
+      get_iter_at_location(iter,x,y);
+      if(iter)
+        get_buffer()->place_cursor(iter);
+      Menu::get().right_click_line_menu->popup(event->button,event->time);
+    } else {
+      Menu::get().right_click_selected_menu->popup(event->button,event->time);
+    }
+    return true;
+  }
   return Gsv::View::on_button_press_event(event);
 }
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,7 +10,6 @@ set(project_files
     entrybox.cc
     info.cc
     juci.cc
-    menu.cc
     notebook.cc
     project.cc
     selectiondialog.cc
@@ -27,6 +26,7 @@ set(project_shared_files
     dispatcher.cc
     filesystem.cc
     git.cc
+    menu.cc
     meson.cc
     project_build.cc
     source.cc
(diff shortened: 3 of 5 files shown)


src/menu.cc
```diff
@@ -226,7 +225,7 @@ Menu::Menu() {
           <attribute name='action'>app.source_goto_method</attribute>
         </item>
         <item>
-          <attribute name='label' translatable='yes'>_Rename</attribute>
+          <attribute name='label' translatable='yes'>_Refactor _Rename</attribute>
```
eidheim: I think refactor is a redundant here. One would say "refactor methods" for instance to represent such methods, but "rename" is well known to be a "refactor method".

src/menu.cc
```diff
@@ -112,6 +112,16 @@ Menu::Menu() {
         </item>
       </section>
       <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Toggle _Line _Comment</attribute>
+        <attribute name='action'>app.source_comments_toggle</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Toggle _Block _Comment</attribute>
```
eidheim: Also not sure if one should move this menu item from the Source menu, as Add Documentation is related (doxygen comments) to Toggle Comments.

src/source.cc
```diff
@@ -2120,7 +2121,25 @@ bool Source::View::on_button_press_event(GdkEventButton *event) {
     get_buffer()->select_range(start, end);
     return true;
   }
-  
+  if((event->type == GDK_BUTTON_PRESS) && (event->button == 3)){
+    auto menu=Menu::get().edit_menu;
```
eidheim: I'm also thinking that we could create a separate menu for the popup on right click, with some selected items from the regular menu (instead of using the edit menu). You can use the same action names as in the regular menu.

eidheim: Actually, maybe two right click popup menus; one for when text is selected, and one for when text is not selected.

zalox: I'm not convinced that two popup menus are the way to go... It could get confusing and in what situations are the menu items different?

I agree that a separate schema for right click is better.

eidheim: I was thinking that cut/copy for instance is not needed in the menu when you right click on a buffer without selected text. And Rename for instance is not needed when you right click on a buffer with selected text.

src/menu.cc
eidheim: In my opinion we should keep it simple. There is added complexity in adding block comments, not only do we have to add more language exceptions, but we have cases where this will result in an error:
```c++
  /*
  /*
  */
  */
```

While line comments always works:
```c++
  /*
  //
  //
  // */
```

Emacs also always add line comments only as far as I know. Finally, you would have one function for adding comments, instead of two.

zalox:
@eidheim Any idea why fedora is failing tests? I made changes, and the end result I like quite well ;)

src/source.cc
```diff
@@ -88,9 +88,83 @@ std::string Source::FixIt::string(Glib::RefPtr<Gtk::TextBuffer> buffer) {
...
+        </item>
+      </section>
+      <section>
+        <item>
+          <attribute name='label' translatable='yes'>_Toggle _Line _Comment</attribute>
+          <attribute name='action'>app.source_comments_toggle</attribute>
+        </item>
+        <item>
+          <attribute name='label' translatable='yes'>_Rename</attribute>
+          <attribute name='action'>app.rename</attribute>
```
eidheim: In order to make it work, I had to change `app.rename` to `app.source_rename` (same as in menu.cc).

src/source.cc
```diff
@@ -2120,7 +2194,25 @@ bool Source::View::on_button_press_event(GdkEventButton *event) {
     get_buffer()->select_range(start, end);
     return true;
   }
-  
+  if((event->type == GDK_BUTTON_PRESS) && (event->button == 3)){
+    if(!get_buffer()->get_has_selection()){
+      int x,y;
+      window_to_buffer_coords(Gtk::TextWindowType::TEXT_WINDOW_WIDGET,event->x,event->y,x,y);
```
eidheim: The x coordinate was somewhat off target. Fixed by using `Gtk::TextWindowType::TEXT_WINDOW_TEXT` instead of `Gtk::TextWindowType::TEXT_WINDOW_WIDGET`.

eidheim:
Looks good, I like it as well:) I'll look further into why the test is failing. 

One thought though: maybe we should move construction of the `right_click_line_menu` and `right_click_selected_menu` (including the xml's) to menu.cc after all? We then have all the menu related code in menu.h/cc. If you do this, to make the tests compile, you would have to move menu.cc from `project_files` to `project_shared_files` in src/CMakeLists.txt. Also, right now, the menus are constructed every time a new source buffer is created.

src/source.cc
```diff
@@ -88,9 +88,83 @@ std::string Source::FixIt::string(Glib::RefPtr<Gtk::TextBuffer> buffer) {
...
+        </item>
+      </section>
+    </menu>
+  </interface>)RAW";
 
 Source::View::View(const boost::filesystem::path &file_path, Glib::RefPtr<Gsv::Language> language): Gsv::View(), SpellCheckView(), DiffView(file_path), language(language), status_diagnostics(0, 0, 0) {
   load();
+  if(!right_click_line_menu || !right_click_selected_menu){
+    auto builder = Gtk::Builder::create_from_string(right_click_menu_xml);
+    const auto create_menu = [=](const std::string &menu){
```
eidheim: Use `[this, builder]` here instead of `[=]`. `[=]` creates copies of every variable available. This might be the cause of the Fedora test fails.

eidheim:
We could also maybe add Go to Declaration, Go to Implementation, and Go to Usage (the most common navigation functions) to the `right_click_line_menu`.

eidheim:
@zalox I played around trying to figure out the test failure, and found that this diff works:
```c++
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index fd17bb6..779a231 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,7 +10,6 @@ set(project_files
     entrybox.cc
     info.cc
     juci.cc
-    menu.cc
     notebook.cc
     project.cc
     selectiondialog.cc
@@ -27,6 +26,7 @@ set(project_shared_files
     dispatcher.cc
     filesystem.cc
     git.cc
+    menu.cc
     meson.cc
     project_build.cc
     source.cc
diff --git a/src/menu.cc b/src/menu.cc
index 9fafdb1..1ae1439 100644
--- a/src/menu.cc
+++ b/src/menu.cc
@@ -3,6 +3,87 @@
 #include <string>
 #include <iostream>
 
+const Glib::ustring right_click_menu_xml= R"RAW(<interface>
+  <menu id='right-click-line-menu'>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Undo</attribute>
+        <attribute name='action'>app.edit_undo</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Redo</attribute>
+        <attribute name='action'>app.edit_redo</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Paste</attribute>
+        <attribute name='action'>app.edit_paste</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Reload _File</attribute>
+        <attribute name='action'>app.reload_file</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Go _to _Declaration</attribute>
+        <attribute name='action'>app.source_goto_declaration</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Go _to _Method</attribute>
+        <attribute name='action'>app.source_goto_method</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Go _to _Usage</attribute>
+        <attribute name='action'>app.source_goto_usage</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Rename</attribute>
+        <attribute name='action'>app.source_rename</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Toggle _Line _Comment</attribute>
+        <attribute name='action'>app.source_comments_toggle</attribute>
+      </item>
+    </section>
+  </menu>
+  <menu id='right-click-selected-menu'>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Undo</attribute>
+        <attribute name='action'>app.edit_undo</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Redo</attribute>
+        <attribute name='action'>app.edit_redo</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Cut</attribute>
+        <attribute name='action'>app.edit_cut</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Copy</attribute>
+        <attribute name='action'>app.edit_copy</attribute>
+      </item>
+      <item>
+        <attribute name='label' translatable='yes'>_Paste</attribute>
+        <attribute name='action'>app.edit_paste</attribute>
+      </item>
+    </section>
+    <section>
+      <item>
+        <attribute name='label' translatable='yes'>_Toggle _Line _Comment</attribute>
+        <attribute name='action'>app.source_comments_toggle</attribute>
+      </item>
+    </section>
+  </menu>
+</interface>)RAW";
+
 Menu::Menu() {
   ui_xml = R"RAW(
 <interface>
@@ -404,10 +485,19 @@ void Menu::build() {
   
   try {
     builder->add_from_string(ui_xml);
-    auto object = Menu::get().builder->get_object("juci-menu");
+    auto object = builder->get_object("juci-menu");
     juci_menu = Glib::RefPtr<Gio::Menu>::cast_dynamic(object);
-    object = Menu::get().builder->get_object("window-menu");
+    object = builder->get_object("window-menu");
     window_menu = Glib::RefPtr<Gio::Menu>::cast_dynamic(object);
+    
+    builder = Gtk::Builder::create();
+    builder->add_from_string(right_click_menu_xml);
+    object = builder->get_object("right-click-line-menu");
+    auto ptr = Glib::RefPtr<Gio::Menu>::cast_dynamic(object);
+    right_click_line_menu = std::make_unique<Gtk::Menu>(ptr);
+    object = builder->get_object("right-click-selected-menu");
+    ptr = Glib::RefPtr<Gio::Menu>::cast_dynamic(object);
+    right_click_selected_menu = std::make_unique<Gtk::Menu>(ptr);
   }
   catch (const Glib::Error &ex) {
     std::cerr << "building menu failed: " << ex.what();
diff --git a/src/menu.h b/src/menu.h
index 9e3fc56..759637c 100644
--- a/src/menu.h
+++ b/src/menu.h
@@ -23,6 +23,9 @@ public:
   Glib::RefPtr<Gio::Menu> juci_menu;
   Glib::RefPtr<Gio::Menu> window_menu;
   
+  std::unique_ptr<Gtk::Menu> right_click_line_menu;
+  std::unique_ptr<Gtk::Menu> right_click_selected_menu;
+  
 private:
   Glib::RefPtr<Gtk::Builder> builder;
   std::string ui_xml;
diff --git a/src/source.cc b/src/source.cc
index 7075694..cbbab43 100644
--- a/src/source.cc
+++ b/src/source.cc
@@ -4,6 +4,7 @@
 #include "terminal.h"
 #include "info.h"
 #include "directories.h"
+#include "menu.h"
 #include <gtksourceview/gtksource.h>
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/spirit/home/qi/char.hpp>
@@ -2123,6 +2124,21 @@ bool Source::View::on_button_press_event(GdkEventButton *event) {
     return true;
   }
   
+  if((event->type == GDK_BUTTON_PRESS) && (event->button == 3)){
+    if(!get_buffer()->get_has_selection()){
+      int x,y;
+      window_to_buffer_coords(Gtk::TextWindowType::TEXT_WINDOW_TEXT,event->x,event->y,x,y);
+      Gtk::TextIter iter;
+      get_iter_at_location(iter,x,y);
+      if(iter)
+        get_buffer()->place_cursor(iter);
+      Menu::get().right_click_line_menu->popup(event->button,event->time);
+    }
+    else
+      Menu::get().right_click_selected_menu->popup(event->button,event->time);
+    
+    return true;
+  }
   return Gsv::View::on_button_press_event(event);
 }
 
diff --git a/src/window.cc b/src/window.cc
index 106efaf..42b1498 100644
--- a/src/window.cc
+++ b/src/window.cc
@@ -32,6 +32,9 @@ Window::Window() {
   configure();
   activate_menu_items();
   
+  Menu::get().right_click_line_menu->attach_to_widget(*this);
+  Menu::get().right_click_selected_menu->attach_to_widget(*this);
+  
   set_default_size(Config::get().window.default_size.first, Config::get().window.default_size.second);
   
   auto directories_scrolled_window=Gtk::manage(new Gtk::ScrolledWindow());
```

Note that this diff is compared to cppit/master.

zalox:
@eidheim I think this is ready for merge.

eidheim:
@zalox Great, and sorry, I missed the last commit, but going to look through and test Tomorrow.

eidheim:
I did some minor changes before merging in https://github.com/cppit/jucipp/commit/fa75b543c7b9da9dfd4150cef97f5b05260f2698:
  * Menu constructor to private (singleton requirement)
  * Removed Reload file from the menu (not sure how common it is to do this, and it is a dangerous function to use since one loose changes in the buffer)
  * Added Go to implementation
  * Slight change in sections and order of menu items (so it's similar to what is found in the normal menu)
  * Hiding tooltips before showing menu

We should maybe consider to remove Undo, Redo, Cut, Paste and Copy from the right click menu as well. These are common operations and I guess most people use shortcuts for these anyway. At least consider removing Undo and Redo as these are not dependent on where you click on the buffer? Not sure though, what do you think?

Anyways, I really like the new right click menus, great job!

eidheim:
I also added https://github.com/cppit/jucipp/commit/e760f9c16f06e67ba500919fcc075b17f437ae1d so that the menus are as similar as possible, that is ordering and naming.

PR #313, opened 2017-02-05
Optional manual auto completion

Istarnion:
I added in an option in config.json for doing manual auto completion, and a keybinding for it.
It was difficult to find a good place to hook up the keybinding to the action. First I tried to get it in the ClangViewAutocomplete constructor, but accelerators are global and I couldn't get them to act on the currently active view. It might be possible, but I don't know GTK well enough.
Second attempt was in the Notebook constructor, but that is called before the config is loaded, so that didn't work either.
In the end, I just put it in juci.cc, in the on_activate function. This is probably not where it should be, but it works now. Any ideas where it might be better suited? Or is there a better way to hook up keybindings all together?

Also, I'm not sure what happened to the submodules.. I think they just got updated. I haven't edited anything in them atleast.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,7 @@
 cmake_minimum_required (VERSION 2.8.8)
 
 project(juci)
-set(JUCI_VERSION "1.2.2.1")
+set(JUCI_VERSION "1.2.2.2")
 
 set(CPACK_PACKAGE_NAME "jucipp")
 set(CPACK_PACKAGE_CONTACT "Ole Christian Eidheim <eidheim@gmail.com>")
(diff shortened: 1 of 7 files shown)


eidheim:
Thank you, this helps a lot. Even though I will probably make another commit based on this PR due to the submodule issue. I'll make you author on the new commit if that is ok. I'll also clean up here and there after.

Istarnion:
Super! Glad I could help

eidheim:
@Istarnion this functionality should now be implemented in https://github.com/eidheim/jucipp/commit/2f6f02670bd064e2f22caa7d956ac873014a2ae2. Could you have a look and test it? I'll wait for your response before merging to cppit/jucipp master.

Istarnion:
I cloned your fork, and it ran exactly as expected. Thanks. Having it as only one option in the config file was alot better :)
I tested the same edge conditions as I did before, and there were no issues.

eidheim:
Thank you again! The commit has been merged. Closing this PR then.

PR #314, opened 2017-02-07
Update install.md

Triangle345:
addec cmake to windows pacman install, just in case. Many people do not have cmake installed by default. Just quality of life.

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -138,7 +138,7 @@ make install
 
 Install dependencies (replace `x86_64` with `i686` for 32-bit MSYS2 installs):
 ```sh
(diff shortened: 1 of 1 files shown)


eidheim:
Thank you, but I think mingw-w64-x86_64-cmake is the correct package. Have you tried? Maybe that resolves #315 as well?

eidheim:
You might need to uninstall the `cmake` package before running cmake, not sure how these two packages link the cmake executable.

eidheim:
The answer is probably that mingw-w64-x86_64-cmake is used when running the MINGW64 shell.

eidheim:
Closing this as I saw #315 was resolved.

Issue #310, opened 2017-01-22
Feature Request: Go to File/Header


donbcd:
It would be very helpful to be able to have a "go to file" or "go to header" function, which I can use on e.g. includes!

adobrich:
Does the "Go to Declaration" (Ctrl+D) & "Go to Implementation" (Ctrl+I) while cursor is over a function not provide the desired result?

eidheim:
It would be nice if one could go to a header when for instance using go to declaration on the header include, for instance having `#include <vector>` one could use go to declaration on the token `vector`. I'll have a look if this is possible.

donbcd:
It comes handy when you do not have a symbol or you do not know what a specific header provides, and still want to take a look at it. This mainly happens with libraries with deeper nested headers, like boost.

eidheim:
Feel free to try the above commit, I'll probably merge it in a few days.

donbcd:
Just tried it and seems to work fine as long as cmake knows about the headers. This is already a big help!

It stops working when digging deeper, e.g. into the headers of a library (I just tried it with gtkmm: the headers which I use in my project work fine, but the headers included inside these headers work only partially, even though they are in the standard paths).

eidheim:
Thank you, I did fix the include regex, that might have been the problem. Also keep in mind that some includes might depend on `#ifdef` for instance.

eidheim:
@donbcd Yet another regex fix: https://github.com/eidheim/jucipp/commit/f538e22f13e264d5eaf89e9473808d4f4a595780

PR #317, opened 2017-02-13
Add handy shortcut: use 'Ctrl - Right Click' to go to implementation directly.

artisdom:
Use  'Ctrl - Right Click' to go to implementation is a very handy feature offered by other IDEs, Jucipp should have it also. :)

### DIFF
--- a/src/source.cc
+++ b/src/source.cc
@@ -2224,7 +2224,15 @@ bool Source::View::on_button_press_event(GdkEventButton *event) {
     get_buffer()->select_range(start, end);
     return true;
   }
-  
+
+  if((event->type == GDK_BUTTON_PRESS) && (event->button == 1)){
+    if(event->state & GDK_CONTROL_MASK) {
+      hide_tooltips();
+      Menu::get().actions["source_goto_implementation"]->activate();
+      return true;
+    }
+  }
+
   if((event->type == GDK_BUTTON_PRESS) && (event->button == 3)){
     hide_tooltips();
     if(!get_buffer()->get_has_selection()){


artisdom:
Sorry, I mean 'Ctrl - Left Click', being working all day, a bit loose in the brain now.

eidheim:
Thank you again! 

This PR require a bit more work it seems. I'll fix it later, hopefully next week or so. The improvements will be as follows:
* Command key should activate this instead of control on MacOS
* Cursor has to be placed I think before activating the menu item
* If clicking on declaration, go to implementation.
* If clicking on implementation, go to declaration.
* If clicking on usage, go to implementation.

artisdom:
Thank you for reviewing my pull request. :)


> Cursor has to be placed I think before activating the menu item

just noticed, the cursor location need to be updated before other actions, otherwise it'll go to the wrong symble.

> If clicking on declaration, go to implementation.
> If clicking on implementation, go to declaration.
> If clicking on usage, go to implementation.

this has more usablity than what i'm proposing.

and it may also touch upon this feature.
[Feature Request: Go to File/Header](https://github.com/cppit/jucipp/issues/310)
Allow user to be able to click and go to header files directly.

eidheim:
I updated the list to include the header files. Will try to fix that issue at the same time.

agauniyal:
This is very good, also highlight with different color or underline to make it look like clicking a link..

eidheim:
The above commit should fix this PR I think. Merging in a day or two.

artisdom:
Just being trying out the new features, I never know how much I love clicking around before, thanks.

PR #318, opened 2017-02-13
Use italic font for comment, more ease on the eyes.

artisdom:
as titled, cheers.


eidheim:
Thank you for your PRs! I'll have a look at them soon.

eidheim:
Maybe I'm used to how it is, but I also see that GitHub for instance uses normal fonts for comments. Keep it as it is, or? I think more people should comment on this, before we close or merge this PR.

artisdom:
While it is customizable through config file, so enjoy it already, cheers.

zalox:
If the user would like italics on comments I suppose he could just edit the theme file in `.juci/styles`.

PR #319, opened 2017-02-15
use canonical file path in Notebook::open to avoid duplicated parsing with soft-linked files.

artisdom:
If there are soft-linked files in the project, the same file is treated as two different files and is parsed mulitple times, and duplicated symbols will show up when doing symbol lookups, like: doing "Go to Method", "Go to Usage".
Using canonical file path in Notebook::open fix the problem.

### DIFF
--- a/src/notebook.cc
+++ b/src/notebook.cc
@@ -130,29 +130,29 @@ std::vector<Source::View*> &Notebook::get_views() {
 }
 
 void Notebook::open(const boost::filesystem::path &file_path, size_t notebook_index) {
+  boost::filesystem::path canonical_path = file_path;
+  if(boost::filesystem::exists(file_path))
+    canonical_path=boost::filesystem::canonical(file_path);
+
   if(notebook_index==1 && !split)
     toggle_split();
   
   for(size_t c=0;c<size();c++) {
-    if(file_path==source_views[c]->file_path) {
+    if(canonical_path==source_views[c]->file_path) {
       auto notebook_page=get_notebook_page(c);
       notebooks[notebook_page.first].set_current_page(notebook_page.second);
       focus_view(source_views[c]);
       return;
     }
   }
-
-  boost::filesystem::path canonical_path = file_path;
   
-  if(boost::filesystem::exists(file_path)) {
-    std::ifstream can_read(file_path.string());
+  if(boost::filesystem::exists(canonical_path)) {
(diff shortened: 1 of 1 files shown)


eidheim:
Great, thank you again, I'll look at this soon.

eidheim:
As it is now, one can open multiple tabs with the same file if one clicks a symbolic link, but they are all the same file. Maybe use canonical_file earler as well (https://github.com/cppit/jucipp/pull/319/files#diff-a58c7d4e7b39ba933684c168a8ab383fL136), so that the symbolic file is fully treated as the canonical file?

artisdom:
You are right. Commit updated to use canonical_file earlier.

eidheim:
After giving this some thought, I have the following proposal: instead of opening the canonical path automatically, one could search for both the normal path and the canonical path when checking if the file is already opened. We would have to store the canonical path of the view in addition, to make sure that clicking on two symbolic links to the same file does not open two tabs.

The reasoning behind the above suggestion is that one should be able to work within a symbolically linked directory and the opened files should look like they are within this directory (not the canonical one). I realise that this would require us to reduce the use of `boost::filesystem::canonical` in other functions as well.

eidheim:
I think we maybe should use `boost::filesystem::absolute` instead of all the occurrences of `boost::filesystem::canonical` in the `src/*` files.

eidheim:
Or not `boost::filesystem::absolute` but instead `boost::filesystem::lexically_normal`. However, sadly `lexically_normal` is not available on all the platforms we support yet.

eidheim:
We would have to create lexically_normal in filesystem.* for the time being.

eidheim:
I think I'll have a look at this later since it seems to require some work and one have to be careful not to break anything.

eidheim:
I fixed the symbolic link issues as described in this PR and in my above comments in the above commit. I will test it more thoroughly before merging. Could you test it also @artisdom and see if it works well on your use cases?

artisdom:
find one issue with starting juci from the terminal,
juci .
juci ..
is not working as before, 
all others looks good.

eidheim:
Thank you for testing! I'll fix the above issue later Tonight.

eidheim:
This issue should be fixed in the latest commits. Thank you for contributing valuable input and help!

PR #320, opened 2017-02-16
Adding a menu item to toggle Tabs visibility

Physiix:
Tested it with Ubuntu 16.04. (Can test it with a Mac if needed)


eidheim:
Thank you, merging in a few minutes

agauniyal:
How do we switch tabs atm? Is there tab cycle key or jump to files?

eidheim:
@agauniyal See Next and Previous Tab in the Window menu. You can configure these in Preferences.

PR #326, opened 2017-03-30
Implemented full screen feature

ingmarlehmann:


eidheim:
Besides some minor changes, I connected to the signal_window_state_event to identify when fullscreen is set or not. It is possible in some window managers to (un)fullscreen applications, and one thus have to track these changes. It did not work entirely well in MacOS, but I guess it's working on Linux at least.

eidheim:
It didn't work 100% in Manjaro either sadly.

eidheim:
I think the above commit fixes the issues with identifying if the window is already fullscreen or not. Merging if everything is ok on MacOS and Manjaro at least.

ingmarlehmann:
Well done! Minor detail but the keybinding is defaulted to nothing now. Perhaps there is a good reason for it but if it is unintentional, I would dare say F11 is the most common keybinding for full screen mode.

eidheim:
I forgot to mention this. The reason is that IDEs have quite a few keybindings compare to regular applications, and only the most used functions should have a default keybinding or else we would fast not have keybinding slots for important new functions. Also, we have to consider many OS'es, and the F keys for instance are not regularly used in MacOS. My F11 for instance is already bound to some OS function and cannot be overwritten by an application. Though Chrome have the following keybinding for fullscreen on MacOS: cmd-control-F, and Chrome for Linux must have some other keybinding for this, maybe F11.

eidheim:
Could maybe use cmd-control-F for MacOS and F11 for other OS'es. I checked a couple applications on Linux and it seems that F11 is quite common.

eidheim:
I added the default shortcuts for full screen discussed above in 25c0b88df204041d029fdf7ad2a82265ecadd723

PR #333, opened 2017-05-13
Add installation instructions for Opensuse

junrrein:
You will notice that in the `cmake` step, I added the option `-DCMAKE_INSTALL_PREFIX=/usr` since, 99% of the time, that's the path you want to use. Unless that's already the default and I haven't noticed.


eidheim:
Thank you. I tested using the OpenSuse Tumbleweed docker image, and had to make some slight changes to the zypper install command. Also I removed the `CMAKE_INSTALL_PREFIX`, and let it be up to the user to add this or use system default. 

Also, the `lldb_test` failed, did not investigate this further though.

PR #336, opened 2017-06-02
add find file option

Joshua-Wright:
Add `Goto File` option under `Source`

It looks like this:
screenshot from 2017-06-01 21-38-41

This is my first open-source code contribution, feedback and/or suggestions welcome!


eidheim:
Thank you! This seems like a nice feature indeed. I'm busy this weekend, but will have a closer look early next week if not before.

src/window.cc
```diff
@@ -656,6 +656,64 @@ void Window::set_menu_actions() {
...
+      auto path_str = filesystem::get_relative_path(path, project_path).string();
+      // SelectionDialog::get()->add_row(path_str.substr(project_path.string().length()+1));
+      SelectionDialog::get()->add_row(path_str);
+    }
+  
+    SelectionDialog::get()->on_select=[this, project_path](const std::string &selected, bool hide_window) {
+      auto full_path = boost::filesystem::canonical(selected);
+      Notebook::get().open(full_path);
+      auto view=Notebook::get().get_current_view();
+      view->hide_tooltips();
```
eidheim: maybe replace with:
```c++
if(auto view=Notebook::get().get_current_view())
  view->hide_tooltips();
```
at least `if(view) view->hide_tooltips();` is important.

src/menu.cc
```diff
@@ -263,6 +263,10 @@ const Glib::ustring menu_xml= R"RAW(<interface>
       </section>
       <section>
         <item>
+          <attribute name='label' translatable='yes'>G_oto File</attribute>
```
eidheim: Replace `G_oto File` with `_Find _File`.

src/window.cc
```diff
@@ -656,6 +656,64 @@ void Window::set_menu_actions() {
     SelectionDialog::get()->show();
   });
   
+  menu.add_action("source_find_file", [this]() {
+    auto view = Notebook::get().get_current_view();
+    auto project_path = boost::filesystem::canonical(Directories::get().path);
```
eidheim: See https://github.com/cppit/jucipp/blob/master/src/window.cc#L599 on how to find the path in which to search.

src/window.cc
```diff
@@ -656,6 +656,64 @@ void Window::set_menu_actions() {
...
+        build_default_path = boost::filesystem::canonical(build->get_default_path());
+      }      
+      if (is_directory(build->get_debug_path())) {
+        build_debug_path = boost::filesystem::canonical(build->get_debug_path());
+      }
+    }
+  
+    // populate with all files in project
+    for (boost::filesystem::recursive_directory_iterator iter(project_path), end; iter != end; iter++) {
+      auto path = boost::filesystem::canonical(iter->path());
```
eidheim: You can remove all uses of `boost::filesystem::canonical` I think as this is taken care of elsewhere in juCi++ in the functions where this is needed.

eidheim:
Thank you again @Joshua-Wright. I'm currently cleaning up the Config class, and will merge your PR after I'm done adding the keybinding. I'll set <primary>p for this feature, and "" as default keybinding for print.

eidheim:
Alright, I think I got it finalised in the above commit, but I'll do some additional tests and do a second look Tomorrow.

PR #342, opened 2017-08-14
Go to Usage and Rename now includes unopened files

eidheim:
Added parsing of unopened files that might contain the searched for token/cursor. You can control the number of threads used in this parsing in preferences: `source.clang_usages_threads`

Also caches parsed files, and stores these caches in the `.usages_clang` folder under the default build path. 

I'll add tests and possibly do some cleanup before merging this. Testing and feedback is welcome, but note that this is a complex addition, and there might be issues.


eidheim:
@zalox Do you know why AppVeyor fails?

eidheim:
@zalox I'm merging this even though appveyor fails since it is not due to source errors. Hopefully, you will find a way to fix the appveyor build later.

Issue #357, opened 2017-12-13
Question: Support for CUDA? Mercurial?


eyalroz:
(I would add the "question" tag but I don't see it)

* Does jucipp support writing and building CUDA code out-of-the-box?
* README.md mentioned support for Git. Is Mercurial also supported?

Overall, this is a very interesting effort.

eidheim:
Thank you for your feedback. 

Although I could not test this since I do not have a Nvidia graphic card, I added initial CUDA support here: https://github.com/eidheim/jucipp/commit/1bb95beb43e600bfe10247b4d6979abd8d50e76a

I added the following lines to make it work on my system, but this might not be necessary if you have added CUDA somehow in your cmake/meson files:
```cmake
add_definitions(-D__CUDACC__)
include_directories(/opt/cuda/include)
```
Feedback and improvements to the commit is very welcome. 

Regarding Mercurial, we do not (yet) support this. Git integration is achieved through the libgit2 library, I do not know if Mercurial has a similar library.

eidheim:
I updated the CUDA support commit to:
https://github.com/eidheim/jucipp/commit/e18c0089707df5c70d59552fae75ed08a672bbe0

PR #366, opened 2018-02-21
Adds support for OpenCL source-files

doe300:
Adds support for OpenCL similar to the support for CUDA added in https://github.com/cppit/jucipp/issues/357.

Support includes:
- auto completion for OpenCL C 2.0 standard-library functions
- documentation tooltips for OpenCL standard-library functions (as long as documented in the [header](https://github.com/llvm-mirror/clang/blob/master/lib/Headers/opencl-c.h))


eidheim:
Thank you!

PR #370, opened 2018-04-08
add includes necessary to compile

Joshua-Wright:
On my system, it was necessary to add these includes in order to compile. Without this, it doesn't recognize `std::function<>`.

My system is ubuntu 16.04, cmake 3.9.0, g++ 5.4.0.


eidheim:
Thank you!

PR #376, opened 2018-05-03
type tooltips when cursor over auto spec

szymonk211:


eidheim:
This was a good idea, thank you:)

PR #381, opened 2018-05-26
Buildsystem refactor and CMake support reimplementation

DWVoid:


DREAMWORLDVOID:
@eidheim I have reimplemeted the cmake support with Codeblocks Project Generators to get more accurate project information. This PR also contains minor refactoring and reformating[I have no exact idea of the project coding style, this PR contains my best guess]
Tested with CMake 2.8.12, 3.7 and 3.10.

eidheim:
Thank you, but this PR has some issues:
* Please do not style reformat the source files. I'd like to use clang-format at some point, but its support for lambda expressions is still poor. Also, when we will perform style format on the entire project, we will create and use a separate user account for this, so that no author gets credited for the changes done through style formatting. 
* The .idea directory should not be added in your commits
* We use boost.property_tree for xml/json processing
* No need to use polymorphism in the cmake/meson source files. These files are intended to be standalone so that they can be used by other projects. Polymorphish is however implemented in `Project::Build` instead. 
* We do not want to enforce use of a generator like the Codeblocks Project Generator. juCi++ should work on a preconfigured project that for instance have used a different generator.
* There might be other issues as well, but I'm having a hard time reading the changes due to the number of line-changes in this PR.

DREAMWORLDVOID:
I am sorry but where is the clang-format configuration file?

